PR #1, opened 2024-03-05
Pull in code

davydog187:


lib/lua.ex
```diff
@@ -0,0 +1,269 @@
...
+    def inspect(lua, _opts) do
+      concat(["#Lua<functions:", "[", Enum.join(Util.user_functions(lua), ", "), "]>"])
+    end
+  end
+
+  @doc """
+  Initializes a Lua VM sandbox. All library functions are stubbed out,
+  so no access to the filesystem or the execution environment is exposed.
+  """
+  def init(_opts \\ []) do
```
davydog187, line 26: change to new, also add a function `sandbox`

PR #14, opened 2024-03-18
Improve the implementations of get/set. Remove .functions

davydog187:
1. Raise better errors for `Lua.get!/2` and `Lua.set!/3`
2. `Lua.load_lua_file!/2` -> `Lua.load_file!/2`
3. Remove `.functions` from state

### DIFF
--- a/test/lua_test.exs
+++ b/test/lua_test.exs
@@ -58,7 +58,6 @@ defmodule LuaTest do
                Lua.eval!(lua, """
                return foo()
                """)
-
     end
 
     test "loading files with syntax errors returns an error" do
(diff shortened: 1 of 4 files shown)


test/lua_test.exs
```diff
@@ -76,13 +88,13 @@ defmodule LuaTest do
         """
 
       assert_raise Lua.CompilerException, error, fn ->
-        Lua.load_lua_file!(Lua.new(), path)
+        Lua.load_file!(Lua.new(), path)
```
davydog187, line 91: Add a test for loading files with the `.lua` extension. We should allow this and strip it

PR #20, opened 2024-04-11
Implement ~LUA for validating Lua code at compile time

davydog187:
Adds a sigil ~LUA which compiles the passed string and raises a Lua.CompilerException if the code does not compile, returning the original string.

This is useful if you have Lua literals in your codebase that you would like to validate before executing them.

### DIFF
--- a/lib/lua.ex
+++ b/lib/lua.ex
@@ -66,6 +66,23 @@ defmodule Lua do
   end
 
   @doc """
+  Write Lua code that is parsed at compile-time. If the code cannot
+  be lexed and parsed, it raises a `Lua.CompilerException`
+
+      iex> ~LUA(return 2 + 2)
+      "return 2 + 2"
+  """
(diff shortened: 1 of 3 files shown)


lib/lua.ex
```diff
@@ -65,6 +65,23 @@ defmodule Lua do
     |> Enum.reduce(%__MODULE__{state: :luerl.init()}, &sandbox(&2, &1))
   end
 
+  @doc """
+  Write Lua code that is parsed at compile-time. If the code cannot
+  be lexed and parsed, it raises a `Lua.CompilerException`
+
+      iex> ~LUA(return 2 + 2)
```
davydog187, line 72: This isn't right, we want to only accept string literals/heredocs

davydog187:
I didn't mean to merge this 😅 I'll fix in another PR

PR #45, opened 2024-09-08
Change Lua.API.install/1 callback to Lua.API.install/3

davydog187:
Changes the arity of the API, passing the scope and data to the `install/3` callback, allowing the install callback to capture runtime data from APIs, databases, or elsewhere in the API scope for further usage.

Note that this is a **breaking change**, but since we have not released a stable version yet, consumers will need to be mindful.


README.md
```diff
@@ -31,7 +31,18 @@
 
 ## Exposing Elixir functions to Lua
 
-`Lua` provides the `deflua` macro for exposing Elixir functions to Lua
+The simplest way to expose an Elixir function to in Lua is using the `Lua.set!/3` function
```
davydog187, line 34:
```suggestion
The simplest way to expose an Elixir function to Lua is using the `Lua.set!/3` function
```

davydog187: Also make a note that functions exposed this way must return lists, while functions using deflua are automatically wrapped.

Another consideration would be to auto wrap functions passed to set!

PR #53, opened 2024-10-03
Auto wrap functions passed to Lua.set! to receive %Lua{}

### LINKED ISSUE #47: Auto-wrap functions passed to Lua.set!/3
Currently, functions passed to `Lua.set!/3` receive the underlying Luerl state record, rather than our wrapped variant. Moreso, the facilities to make sure that data is passed cleanly are also removed.

In order to support wrapping functions, we must refactor `Lua.load_api` to use a lower-level API, otherwise functions will get double-wrapped and cause more issues

### LINKED ISSUE #51: Confusing return type expected for deflua functions with state
See the following example from @ivarvong https://gist.github.com/ivarvong/3b935ec2325d7ef18031cc58969ba2b7

When `deflua` functions return values, we expect them to return a tuple of `{[list_of_return_values], lua_state}`. If you don't return this, you get a confusing error message.

Make this better

davydog187:
This PR fixes the problem where functions passed to `Lua.set!/3` were receiving the underlying Luerl record rather than a `%Lua{}` struct.

Also provides more flexibility to the return values of `deflua` functions that take Lua state, allowing them to omit passing the state, or return single values, and we will auto wrap them for you

Closes #47
Closes #51

### DIFF
--- a/lib/lua.ex
+++ b/lib/lua.ex
@@ -166,7 +166,28 @@ defmodule Lua do
       iex> {[10], _lua} = Lua.eval!(lua, "return sum(1, 2, 3, 4)")
 
   """
-  def set!(%__MODULE__{state: state}, keys, value) do
+  def set!(%__MODULE__{} = lua, keys, value) do
+    value =
+      case value do
+        func when is_function(func, 1) ->
+          func
+
+        func when is_function(func, 2) ->
+          fn args, state ->
+            case func.(args, wrap(state)) do
+              {value, %__MODULE__{} = lua} -> {List.wrap(value), lua.state}
+              value -> {List.wrap(value), state}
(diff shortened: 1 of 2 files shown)


lib/lua.ex
```diff
@@ -166,7 +166,28 @@ defmodule Lua do
...
+    value =
+      case value do
+        func when is_function(func, 1) ->
+          func
+
+        func when is_function(func, 2) ->
+          fn args, state ->
+            case func.(args, wrap(state)) do
+              {value, %__MODULE__{} = lua} -> {List.wrap(value), lua.state}
+              value -> {List.wrap(value), state}
```
davydog187, line 179: need to handle tables

Issue #77, opened 2025-05-14
Add execution quotas


crertel:
## Idea

Similar to how luaerl [has this](https://hexdocs.pm/luerl/luerl_sandbox.html) has a way of specifying max reductions, it'd be neat to have that ability to limit both the max reductions (in BEAM-land) or the max lua bytecode instructions.

## Suggestions

* Add a `max_reductions` or `max_instructions`  keys to `Lua.eval!`'s `options` param.
* Change the return arity in such cases to 3: `{:done, ...returned values..., ...returned state...}` or `{:exceeded_reductions, ...returned state...}` or `{:exceeded_instructions, ...returned state...}`

## Prior art

* https://hexdocs.pm/luerl/luerl_sandbox.html

crertel:
> However, if we wanted to simply do the max_reductions, we could build a process based sandbox like Luerl has.

So, this is a really good point.

> The benefit of this approach is that it could lead to more interesting ways of sandboxing and open up new capabilities. The downside is that it must be implemented in Luerl and is orders of magnitude more complicated.

Also very good points.

I think maybe there's two different things we're kinda feeling out:

* Sometimes, we want to say "look friend just run the code and kill it if it would take too long".
* Sometimes, we want to say "okay just keep executing this much, and then we're gonna force you to pre-emptively multitask from the execution environment. you can be as gross as you want, but we'll keep you from hogging the system."

The first one is "easy", the second is more useful but harder.

Maybe an incremental approach? Like, just the sandbox itself would be kinda handy, and then we buy Robert sufficient beer to have a stab at the other one?

crertel:
The semantics also get kinda weird if, for example, the user runs a process that calls back into Elixir or whatever and that has side-effects--the sandbox dies and doesn't really "return" anything, but maybe touched stuff outside its lifecycle. I'd argue that's just something to warn users about.

davydog187:
Once a function calls into Elixir or Erlang, all bets are off on pausing execution. I'm not aware of a BEAM mechanism for pausing and resuming a process from where it left off.

The process based sandbox is really easy, as you mentioned the more sophisticated version that yields back will require a lot more work in Luerl itself.

I am happy to work on those changes to Luerl, I will need to get buy in from Robert first

crertel:
That's awesome! And yeah, good call on callbacks being unreliable.

KelvinHu:
You got me there, I didn't think that far ahead.

davydog187:
[...]

PR #78, opened 2025-05-14
Stub in max_reductions, max_instructions opts and tests.

crertel:
Just a starting point for #77 !

I saw that this starts to get into luerl, and I don't really know that well enough to touch what probably needs to happen next.

### DIFF
--- a/test/lua_test.exs
+++ b/test/lua_test.exs
@@ -395,6 +395,38 @@ defmodule LuaTest do
                return global, success, message
                """)
     end
+
+    test "it can be given a max_reductions option" do
+      assert {:done, ["done"], _} =
+               Lua.eval!("return \"done\" ", max_reductions: :infinity)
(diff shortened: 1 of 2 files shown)


test/lua_test.exs
```diff
@@ -395,6 +395,38 @@ defmodule LuaTest do
...
+               Lua.eval!(
+                 """
+                 local a = 1
+                 b = a + 1
+                 c = b + 1
+                 return "done"
+                 """,
+                 max_instructions: 1
+               )
+    end
```
crertel, line 429: These currently break as we don't have the correct return value yet--but, stubbing that in would add too many hacks for the moment so I just left these ready.

davydog187:
@crertel I am very open to adding more sandboxing features, but I am going to close this until this is ready for more work.

PR #84, opened 2025-06-04
feat(table): function for choosing how to encode empty table values

### LINKED ISSUE #83: Deepcast empty table to empty list
Hi,
I've come accross the need to deepcast a nested LUA table that might contain empty tables which should be interpreted either as empty lists or empty maps. There does not seem to be a way to make a distinction yet since the `deepcast` heuristic does not work on empty table.

For example I'd like to be able to write something along the lines of
```elixir
{_, env} = Lua.new() |> Lua.eval!("my_object = {my_empty_list = empty_list(), my_empty_map = {}}")
table = Lua.get!(env, [:my_object])
Lua.decode!(env, table) |> Lua.Table.deep_cast()
# %{"my_empty_list" => [], "my_empty_map" => %{}}
```

One option I see would be to inject a special value via the call to `empty_list` that gets pattern matched to an empty list in `deep_cast` but maybe there is a better way?

Thanks !

davydog187:
Closes #83

### DIFF
--- a/lib/lua/table.ex
+++ b/lib/lua/table.ex
@@ -157,19 +157,47 @@ defmodule Lua.Table do
 
       iex> Lua.Table.deep_cast([{"a", 1}, {"b", [{1, 3}, {2, 4}]}])
       %{"a" => 1, "b" => [3, 4]}
+
+      iex> Lua.Table.deep_cast([{"a", []}], empty_as: :map)
+      %{"a" => %{}}
+
+      iex> Lua.Table.deep_cast([{"a", []}], empty_as: :list)
+      %{"a" => %{}}
+
+  ## Opts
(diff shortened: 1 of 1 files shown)


lib/lua/table.ex
```diff
@@ -157,19 +157,49 @@ defmodule Lua.Table do
...
+  keys are passed as nil. Defaults to a function that returns maps
   """
-  def deep_cast(value) do
+  def deep_cast(value, opts \\ []) do
+    opts = Keyword.validate!(opts, empty_encoder: fn _ -> %{} end)
+
+    do_deep_cast(value, Keyword.fetch!(opts, :empty_encoder))
+  end
+
+  defp do_deep_cast(value, empty_encoder) do
```
davydog187, line 177: I need to handle top level keys / maps

davydog187:
Will come back to this later.

PR #88, opened 2025-08-22
feat: bump luerl to 1.5.0 (#86)

### LINKED ISSUE #86: Depend on `{:luerl, "<~ 1.5.0"}`
luerl has had a new version published :)

bu6n:
This bumps the luerl dependency to 1.5.0 and makes tests pass again.

Failing tests were either due to:
- unicode Handling (this PR basically reverses 598e0ae because it seems now handled correctly by luerl)
- changed error messages

The [changelog](https://github.com/rvirding/luerl/blob/develop/CHANGELOG.md#150) mentions other things as well, but I'm not sure if these influence this library or not. 

closes #86

### DIFF
--- a/lib/lua.ex
+++ b/lib/lua.ex
@@ -285,11 +285,19 @@ defmodule Lua do
     end
   end
 
-  defp illegal_index([:_G | keys]), do: illegal_index(keys)
   defp illegal_index(["_G" | keys]), do: illegal_index(keys)
 
   defp illegal_index(keys) do
-    {:illegal_index, nil, Enum.join(keys, ".")}
+    {:illegal_index, nil,
+     keys
+     |> Enum.map(fn key ->
+       if(is_tuple(key)) do
+         inspect(key)
+       else
+         key
+       end
+     end)
+     |> Enum.join(".")}
   end
 
   @doc """


davydog187:
Thank you for the PR! Running the tests now

bu6n:
I think there might have been some problem with builds.hex.pm right when running the builds. Should they be retried?

bu6n:
There seem to be a dialyzer check that's not passing. It's related to `lib/lua.ex:288`:

```elixir
  defp illegal_index([:_G | keys]), do: illegal_index(keys)
  defp illegal_index(["_G" | keys]), do: illegal_index(keys)

  defp illegal_index(keys) do
    {:illegal_index, nil, Enum.join(keys, ".")}
  end
```

It seems that `:_G` cannot match anymore:
```
lib/lua.ex:288:8:pattern_match
The pattern can never match the type.

Pattern:
[:_G | _keys]

Type:
[
  false
  | nil
  | true
  | binary()
  | number()
  | {:eref, _}
  | {:erl_func, _}
  | {:tref, _}
  | {:usdref, _}
  | {:funref, _, _}
  | {:erl_mfa, _, _, _}
]
```

Do you think it is safe to just remove the line 288?

davydog187:
For reference:

```elixir
 {1, :luerl_scan, {:user, ~c"syntax error near '<\\34>'"}}
```

davydog187:
PR sent upstream to Luerl https://github.com/rvirding/luerl/pull/221

smn:
thank you!

PR #117, opened 2026-02-09
feat: String standard library

davydog187:
## Summary

Implements the string.* functions from Lua 5.3, providing comprehensive string manipulation capabilities.

### Simple String Operations
- `string.lower(s)` - converts string to lowercase
- `string.upper(s)` - converts string to uppercase  
- `string.len(s)` - returns length of string
- `string.reverse(s)` - reverses a string
- `string.rep(s, n [, sep])` - repeats string n times with optional separator

### Substring Operations
- `string.sub(s, i [, j])` - extracts substring from position i to j
  - Supports Lua's 1-based indexing
  - Supports negative indices (from end)
  - Returns empty string for out-of-range indices

### Byte Operations
- `string.byte(s [, i [, j]])` - returns byte values of characters
  - Can return multiple values for a range
  - Supports negative indices
- `string.char(...)` - creates string from byte values
  - Validates bytes are in 0-255 range

### String Formatting
- `string.format(fmt, ...)` - C-style format strings
  - `%s` - string
  - `%d`, `%i` - integer
  - `%f` - float (6 decimal places)
  - `%x`, `%X` - hexadecimal (lower/upper case)
  - `%o` - octal
  - `%c` - character from byte value
  - `%q` - quoted string with escaped special characters
  - `%%` - literal percent sign

## Implementation Details

- String library registered as global `string` table
- All functions validate argument types with clear error messages
- Proper Lua 1-based indexing with negative index support
- Compatible with Lua 5.3 semantics

## Pattern Matching Deferred

The following functions are deferred to a future PR as they require implementing Lua's pattern matching engine:
- `string.find(s, pat, init, plain)`
- `string.match(s, pat, init)`
- `string.gmatch(s, pat)` 
- `string.gsub(s, pat, repl, n)`

## Tests

32 comprehensive tests covering:
- All string functions with valid inputs
- Edge cases (empty strings, out-of-range indices, zero repetitions)
- Type validation and error handling
- All format specifiers
- String table accessibility

## Changes

- Added `lib/lua/vm/stdlib/string.ex` - Complete string library implementation
- Modified `lib/lua/vm/stdlib.ex` - Register string library on install
- Added `test/lua/vm/string_test.exs` - 32 comprehensive tests

## Test Results

✅ All **980 tests passing** (32 new tests added)

### DIFF
--- a/lib/lua/vm/argument_error.ex
+++ b/lib/lua/vm/argument_error.ex
@@ -36,7 +36,7 @@ defmodule Lua.VM.ArgumentError do
         details: "value out of range"
   """
 
-  defexception [:function_name, :arg_num, :expected, :got, :details, :message]
+  defexception [:function_name, :arg_num, :expected, :got, :details]
 
   @impl true
   def exception(opts) do
@@ -46,16 +46,24 @@ defmodule Lua.VM.ArgumentError do
     got = Keyword.get(opts, :got)
     details = Keyword.get(opts, :details)
 
-    message = build_message(function_name, arg_num, expected, got, details)
-
     %__MODULE__{
       function_name: function_name,
       arg_num: arg_num,
       expected: expected,
       got: got,
-      details: details,
-      message: message
+      details: details
     }
+  end
+
+  @impl true
+  def message(%__MODULE__{
+        function_name: function_name,
+        arg_num: arg_num,
+        expected: expected,
+        got: got,
+        details: details
+      }) do
+    build_message(function_name, arg_num, expected, got, details)
   end
 
   defp build_message(function_name, arg_num, expected, got, details) do


lib/lua/vm/stdlib/string.ex
```diff
@@ -0,0 +1,391 @@
...
+  defp string_rep([str, n | _], _state) when is_binary(str) do
+    raise RuntimeError, value: "bad argument #2 to 'rep' (number expected, got #{typeof(n)})"
+  end
+
+  defp string_rep([other, _ | _], _state) do
+    raise_string_expected(1, "rep", other)
+  end
+
+  defp string_rep([_], _state) do
+    raise RuntimeError, value: "bad argument #2 to 'rep' (number expected)"
```
davydog187: whats up with #2 here? should this be hardcoded?

lib/lua/vm/stdlib/string.ex
```diff
@@ -0,0 +1,391 @@
+defmodule Lua.VM.Stdlib.String do
+  @moduledoc """
+  String standard library functions for Lua.
+
+  Implements the string.* functions from Lua 5.3.
```
davydog187: i'd love to see much better documentation here since this part will go to hex docs. we should probably mention which functions this module is exposing

PR #118, opened 2026-02-09
feat(phase-4): Type-Safe Arithmetic & Comparisons

davydog187:
### Goal
Proper Lua 5.3 arithmetic semantics with clear error messages.

### Implementation

**Float Division by Zero**: Updated to approximate Lua 5.3 behavior:
- `positive / 0` returns `1.0e308` (approximating +inf)
- `negative / 0` returns `-1.0e308` (approximating -inf)  
- `0 / 0` returns `0.0` (approximating NaN)

Note: True IEEE 754 infinity/NaN are difficult to create in Elixir/Erlang due to strict arithmetic checks. This is a documented limitation.

**Arithmetic Type Checking** (already complete):
- All arithmetic operations use `safe_*` helpers with type checking
- `TypeError` raised for non-numeric operands with clear messages
- String-to-number coercion follows Lua 5.3 semantics

**Comparison Safety** (already complete):
- `==` and `~=` work on any types (return false for different types)
- `<`, `<=`, `>`, `>=` only work on numbers or strings (same type)
- `TypeError` raised for incompatible type comparisons
- String comparison is lexicographic

### Changes
- `lib/lua/vm/executor.ex` - Updated `safe_divide` to handle division by zero
- `test/lua/vm/arithmetic_test.exs` - Updated division by zero tests

### Verification
- [x] `mix format` completed
- [x] `mix compile --warnings-as-errors` passes
- [x] `mix test --exclude pending` passes (1,011 tests)
- [x] All arithmetic type checking tests pass
- [x] All comparison safety tests pass

### DIFF
--- a/lib/lua/vm/executor.ex
+++ b/lib/lua/vm/executor.ex
@@ -935,13 +935,20 @@ defmodule Lua.VM.Executor do
   defp safe_divide(a, b) do
     with {:ok, na} <- to_number(a),
          {:ok, nb} <- to_number(b) do
-      # Check for division by zero
-      # Note: Standard Lua 5.3 returns inf/-inf/nan for float division by zero,
-      # but Elixir doesn't support creating these values easily, so we raise an error
-      if nb == 0 or nb == 0.0 do
-        raise Lua.VM.RuntimeError, value: "attempt to divide by zero"
-      else
-        na / nb
+      # Float division by zero returns inf/-inf/nan (Lua 5.3 behavior)
+      cond do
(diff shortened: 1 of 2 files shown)


davydog187:
I'm not sure that we want to do this, I think we should potentially diverge from Lua 5.3 in this case

davydog187:
Closing — superseded by #177 (A0: 64-bit integer overflow wrapping). The divide-by-zero approximation here (returning 1.0e308 as fake +inf) is no longer the direction we want; current arithmetic semantics live in the post-A0 executor.

PR #130, opened 2026-02-11
Implement full varargs support for Lua 5.3

davydog187:
## Summary
- Implements comprehensive varargs (`...`) support for Lua 5.3 compatibility
- Enables running Lua 5.3 test suite files like `vararg.lua`
- All varargs test cases pass (7/7)

### Compiler
- **Return statements**: Detects `return a, b, ...` and generates vararg expansion with variable return count
- **Table constructors**: Detects `{a, b, ...}` and generates vararg expansion for list fields
- **Function calls**: Detects `f(a, b, ...)` and generates vararg expansion with encoded arg counts
- Uses negative arg counts in call instructions to encode fixed + vararg arguments

### VM
- **vararg instruction**: Supports `count==0` to load all varargs into consecutive registers
- **return instruction**: Supports `count==-1` to return all values including expanded varargs
- **set_list instruction**: Supports `count==0` for variable number of values from varargs
- **call instruction**: Decodes negative arg counts as fixed args + varargs

## Features Supported
- ✅ Simple varargs: `function f(...) return ... end`
- ✅ Mixed params: `function f(a, b, ...) return a, b, ... end`
- ✅ Varargs in tables: `{...}` and `{a, b, ...}`
- ✅ Varargs in function calls: `g(...)` and `g(a, b, ...)`
- ✅ Empty varargs handling
- ✅ Works with `select('#', ...)` and other stdlib functions

## Test Results
```
7 varargs tests, 0 failures
Full test suite: 1094 tests, 0 failures
```

### DIFF
--- a/lib/lua/compiler/codegen.ex
+++ b/lib/lua/compiler/codegen.ex
@@ -111,7 +111,7 @@ defmodule Lua.Compiler.Codegen do
         {value_instructions, result_reg, ctx} = gen_expr(value, ctx)
         {value_instructions ++ [Instruction.return_instr(result_reg, 1)], ctx}
 
-      multiple when length(multiple) > 1 ->
+      [_, _ | _] = multiple ->
         # Check if last value is vararg - needs special handling
         {init_values, last_value} = Enum.split(multiple, -1)
         [last] = last_value
@@ -129,9 +129,11 @@ defmodule Lua.Compiler.Codegen do
                 {value_instructions, value_reg, ctx} = gen_expr(value, ctx)
 
                 move =
-                  if value_reg == target_reg,
-                    do: [],
-                    else: [Instruction.move(target_reg, value_reg)]
+                  if value_reg == target_reg do
+                    []
+                  else
+                    [Instruction.move(target_reg, value_reg)]
+                  end
 
                 {instructions ++ value_instructions ++ move, ctx}
               end)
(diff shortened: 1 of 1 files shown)


lib/lua/compiler/codegen.ex
```diff
@@ -111,23 +111,60 @@ defmodule Lua.Compiler.Codegen do
         {value_instructions, result_reg, ctx} = gen_expr(value, ctx)
         {value_instructions ++ [Instruction.return_instr(result_reg, 1)], ctx}
 
-      multiple ->
-        base_reg = ctx.next_reg
+      multiple when length(multiple) > 1 ->
```
davydog187: use pattern matching

lib/lua/compiler/codegen.ex
```diff
@@ -111,23 +111,60 @@ defmodule Lua.Compiler.Codegen do
...
+              init_values
+              |> Enum.with_index()
+              |> Enum.reduce({[], ctx}, fn {value, i}, {instructions, ctx} ->
+                target_reg = base_reg + i
+                {value_instructions, value_reg, ctx} = gen_expr(value, ctx)
+
+                move =
+                  if value_reg == target_reg,
+                    do: [],
+                    else: [Instruction.move(target_reg, value_reg)]
```
davydog187: we should integrate styler here

PR #142, opened 2026-02-14
feat(phase-17): 64-bit integer overflow wrapping and FuncDecl local assignment

davydog187:
## Summary

Critical fixes for Lua 5.3 integer semantics and function declaration behavior.

### 64-bit Integer Overflow Wrapping ⭐
All integer operations now wrap to signed 64-bit per Lua 5.3 spec:
- **Bitwise operations**: `<<`, `>>`, `&`, `|`, `~`, `^` wrap to signed 64-bit
- **Arithmetic operations**: `+`, `-`, `*`, `//`, `%`, unary `-` wrap for integer operands
- Added `to_signed_int64/1` helper for proper IEEE 754 compliance
- `maxint + 1` now correctly equals `minint`
- `1 << 63` now correctly equals `minint` (not overflow to big integer)

### FuncDecl Local Assignment ⭐
Fixed `function f()` syntax to properly update local variables:
- `function f()` now correctly updates local `f` when one exists in scope
- Scope resolution marks assignment target in `var_map` at declaration time
- Prevents future locals from incorrectly affecting earlier code during codegen
- Fixes common Lua pattern where `local f = function()...` is followed by `function f()`

**Example fixed pattern:**
```lua
local f = function(x) return "first" end
-- Later redefine f
function f(x) return "second" end
-- f now correctly refers to "second"
```

## Implementation Details

**Integer Wrapping:**
- `to_signed_int64/1` masks to 64 bits then converts to signed range
- Applied to all bitwise ops: `band`, `bor`, `bxor`, `bnot`, `bsl`, `bsr`
- Applied to integer arithmetic: `safe_add`, `safe_subtract`, `safe_multiply`, `safe_negate`, `lua_idiv`, `safe_modulo`
- Float operations remain IEEE 754 (no wrapping)

**FuncDecl Fix:**
- Scope resolution checks if local exists at FuncDecl position
- Stores target as `{:func_decl_target, decl}` in `var_map`
- Codegen uses this marker instead of checking `ctx.scope.locals` (which contains all function-scope locals)

## Test Status
- **Unit tests**: 1,273 passing, 0 failures ✅
- **No regressions**: All existing tests continue to pass
- **Enables progress**: constructs.lua now progresses past line 160 (was failing at line 20)

## Dependencies
Depends on #141 (vararg and scope fixes).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #152, opened 2026-02-18
fix: local function declarations

### LINKED ISSUE #144: local function declarations are silently dropped
**Severity**: High
**Area**: Compiler — `codegen.ex`

### Description

`local function name() ... end` (i.e., `Statement.LocalFunc` AST nodes) has no codegen handler. The catch-all `defp gen_statement(_stmt, ctx), do: {[], ctx}` silently discards the declaration — no instructions are emitted and no error is raised.

### Reproduction

```lua
local function add(a, b)
  return a + b
end
return add(1, 2)
-- Expected: 3
-- Actual: runtime error — "add" is nil
```

### Workaround

Use a regular (global) function declaration or an assignment:

```lua
function add(a, b)   -- global
  return a + b
end

local add = function(a, b)   -- explicit assignment
  return a + b
end
```

davydog187:
Pulled from #142
Closes #144 

### FuncDecl Local Assignment ⭐
Fixed `function f()` syntax to properly update local variables:
- `function f()` now correctly updates local `f` when one exists in scope
- Scope resolution marks assignment target in `var_map` at declaration time
- Prevents future locals from incorrectly affecting earlier code during codegen
- Fixes common Lua pattern where `local f = function()...` is followed by `function f()`

**Example fixed pattern:**
```lua
local f = function(x) return "first" end
-- Later redefine f
function f(x) return "second" end
-- f now correctly refers to "second"
```

**FuncDecl Fix:**
- Scope resolution checks if local exists at FuncDecl position
- Stores target as `{:func_decl_target, decl}` in `var_map`
- Codegen uses this marker instead of checking `ctx.scope.locals` (which contains all function-scope locals)

### DIFF
--- a/lib/lua/compiler/codegen.ex
+++ b/lib/lua/compiler/codegen.ex
@@ -464,9 +464,9 @@ defmodule Lua.Compiler.Codegen do
     {[loop_instruction], ctx}
   end
 
-  defp gen_statement(%Statement.ForNum{var: var, start: start_expr, limit: limit_expr, step: step_expr, body: body}, ctx) do
-    # Get the loop variable's register from scope
-    loop_var_reg = ctx.scope.locals[var]
+  defp gen_statement(%Statement.ForNum{start: start_expr, limit: limit_expr, step: step_expr, body: body} = node, ctx) do
+    # Get the loop variable's register from var_map (locals are restored after scope isolation)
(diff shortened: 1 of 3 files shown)


davydog187:
Closing — superseded by #185 ("fix(compiler): assign function declaration to in-scope local"), which already closed #144 on main.

PR #179, opened 2026-05-05
test(A1): pin missing-key table read behaviour

### LINKED ISSUE #162: Empty/missing-key table reads return nil instead of crashing
Tracked by plan: [`.agents/plans/A1-table-nil-on-missing-key.md`](https://github.com/tv-labs/lua/blob/main/.agents/plans/A1-table-nil-on-missing-key.md)

**Status**: `ready`
**Direction**: A — Suite Triage

When this plan is picked up via `/next-plan`:

2. Plan status moves `ready → in-progress → review`.
3. A PR is opened with `Closes #<this-issue>` in the body.
4. Merge closes this issue.

davydog187:
## A1: Empty/missing-key table reads return nil instead of crashing

Plan: [`.agents/plans/A1-table-nil-on-missing-key.md`](../blob/fix/table-nil-on-missing-key/.agents/plans/A1-table-nil-on-missing-key.md)
Closes #162

### Goal

Reading a missing key from a table must return `nil`, not raise
`Map.fetch!` (which currently surfaces as `Lua runtime error: key N not
found in: %{}`).

### TL;DR

**A1 already passes in current code.** The fix appears to have landed
earlier — most likely as part of the CPS executor refactor in PR #156.
This PR adds regression coverage so it cannot silently come back, and
reroutes the actually-still-broken slice of A1's "unlocks" list to a new
plan **A15** (different root cause, same exception class).

No production code changes here. Just tests + plan-file housekeeping.

### Success criteria

- [x] `mix test` passes (1284 → 1294, +10 new tests, 0 failures)
- [x] New unit tests in `test/lua/vm/table_index_test.exs`:
  - [x] `local t = {}; return t[5]` → `[nil]`
  - [x] `local t = {1,2,3}; return t[10]` → `[nil]`
  - [x] `local t = {a=1}; return t.b` → `[nil]`
  - [x] `local t = {}; return t[5] == nil` → `[true]`
  - [x] Metatable `__index` (function form, table form, direct-hit short-circuit) — regression guard
  - [x] Bonus: `rawget`, `next({})`, `pairs({})` cases
- [x] `mix test --only lua53` count unchanged (29 tests, 0 failures, 25 skipped). The unlock-list files are blocked by **other** bugs (see Discoveries) and are now tracked separately.

### Discoveries

The four explicit success-criteria cases plus metatable `__index`,
`rawget`, `pairs`/`ipairs`/`next` over empty tables all already returned
`nil` correctly when this plan was picked up.

The error message A1 describes (`key N not found in: %{}`) does still
appear in `sort.lua`, `strings.lua`, `verybig.lua`, but the source is
**not** a missing `table.data` key. It is `Map.fetch!(state.open_upvalues, reg)`
in `lib/lua/vm/executor.ex` at lines 301 and 310 — an open-upvalue
tracking bug that surfaces with the same exception class. That work is
now plan **A15** (`fix/open-upvalue-missing-cell`).

`nextvar.lua` fails for a different reason (`attempt to concatenate a
nil value`), reproducer points at the same area as the for-loop register
regression covered by **A14**.

### Changes

```
 .agents/plans/A1-table-nil-on-missing-key.md  |  +44 -2
 .agents/plans/A15-open-upvalue-missing-cell.md |  +new (134 lines)
 test/lua/vm/table_index_test.exs               |  +new (157 lines)
```

### Verification

```
mix format
mix test
# 52 doctests, 45 properties, 1294 tests, 0 failures, 32 skipped

mix test --only lua53
# 29 tests, 0 failures, 25 skipped (1362 excluded)
```

- Metatable `__index` resolution (already works; only added regression coverage).
- Performance optimization of table access.
- The open-upvalue crash that surfaces in some unlock-list files (now plan A15).
- The for-loop register regression in nextvar.lua (already covered by plan A14).

### DIFF
--- a/test/lua/vm/table_index_test.exs
+++ b/test/lua/vm/table_index_test.exs
@@ -7,15 +7,17 @@ defmodule Lua.VM.TableIndexTest do
   alias Lua.VM.State
   alias Lua.VM.Stdlib
 
-  # Regression coverage for the bug described in A1: reading a missing key
-  # from a table must return nil, not crash with `Map.fetch!` /
-  # `key N not found in: %{}`.
+  # Reading a missing key from a table must return nil, never raise. This
+  # covers the language-level contract from Lua 5.3 §3.4.7: an indexing
+  # access `t[k]` for a key that is not present in `t` evaluates to nil
+  # (after consulting `__index` if a metatable is set).
   #
-  # The cases below were the explicit success criteria for plan A1. By the
-  # time A1 was investigated they all already passed (the underlying fix
-  # had landed in earlier work, most likely the CPS executor refactor in
-  # PR #156). These tests pin the behaviour so it cannot regress silently
-  # again, and document what "table read returns nil" means in this VM.
+  # The cases below pin that contract against silent regression on the
+  # most common shapes: empty tables, out-of-bounds array reads, missing
+  # string fields, comparisons with nil, metatable `__index` fall-through
+  # in both function and table form, the "direct hit must not consult
+  # __index" short-circuit, and the stdlib helpers (`rawget`, `next`,
+  # `pairs`) that all share the same lookup path.
 
   describe "missing key reads return nil" do
     test "empty table returns nil for any integer key" do
--- a/test/lua/vm/table_index_test.exs
+++ b/test/lua/vm/table_index_test.exs
@@ -1,0 +1,162 @@
+defmodule Lua.VM.TableIndexTest do
+  use ExUnit.Case, async: true
+
+  alias Lua.Compiler
+  alias Lua.Parser
+  alias Lua.VM
+  alias Lua.VM.State
+  alias Lua.VM.Stdlib
+
+  # Regression coverage for the bug described in A1: reading a missing key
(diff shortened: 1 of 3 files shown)


test/lua/vm/table_index_test.exs
```diff
@@ -0,0 +1,162 @@
...
+
+  # Regression coverage for the bug described in A1: reading a missing key
+  # from a table must return nil, not crash with `Map.fetch!` /
+  # `key N not found in: %{}`.
+  #
+  # The cases below were the explicit success criteria for plan A1. By the
+  # time A1 was investigated they all already passed (the underlying fix
+  # had landed in earlier work, most likely the CPS executor refactor in
+  # PR #156). These tests pin the behaviour so it cannot regress silently
+  # again, and document what "table read returns nil" means in this VM.
```
davydog187: Let's avoid referring to ephemeral resources in the plan files like "A1". This does not have good longevity

davydog187: Good call. Rewrote the header in de1df74 to describe the Lua 5.3 §3.4.7 contract the tests pin, without naming plans or PR numbers.

PR #193, opened 2026-05-06
Perf/cps executor

davydog187:
not sure if we want anything in here, i had this stashed locally


davydog187:
Closing — stashed WIP that's now stale and conflicting. Will revisit ideas in Direction B (perf investigation) where appropriate.

PR #202, opened 2026-05-07
fix(vm): track dead keys so pairs survives mid-iteration deletion

davydog187:
## Dead-key tracking for `next` so `pairs` survives mid-iteration deletion

Plan: [`.agents/plans/A7a-nextvar-dead-keys.md`](https://github.com/tv-labs/lua/blob/fix/nextvar-dead-keys/.agents/plans/A7a-nextvar-dead-keys.md)

### Goal

Lua 5.3 §6.1 says `next(t, k)` is well-defined while `pairs` is iterating
even if the body just cleared `t[k]`. The reference VM keeps the cleared
key reachable in the hash chain (`TDEADKEY`) so iteration finds the next
live entry. The same paragraph also requires `next(t, k)` to raise
"invalid key to 'next'" when `k` was never a key in `t`.

This PR adds the bookkeeping that supports both contracts and routes
every table-write site through it.

### Success criteria

- [x] `mix test` passes (1420 → 1430, no regressions, +10 unit tests).
- [x] `next(t, k)` raises `bad argument #2 to 'next' (invalid key to 'next')`
      when `k` was never a key in `t` (Lua 5.3 §6.1; nextvar.lua line 230).
- [x] `for k, v in pairs(t) do t[k] = nil end` iterates every key once
      without raising (nextvar.lua line 320–326).
- [x] Unit tests in `test/lua/vm/nextvar_dead_keys_test.exs` cover:
      strict-`next` (4 cases), iterate-then-clear (3 cases), dead-key
      revival without state leaks (3 cases). 10 total.
- [ ] `nextvar.lua` passes end-to-end and is promoted to `@ready_tests` —
      **deferred**. With dead-key tracking landed, bisecting `nextvar.lua`
      from a clean slate revealed two pre-existing semantic gaps further
      down the file (table-library metamethods at 389–433; numeric-`for`
      string coercion at 510). Both verified to fail on `main` before this
      PR. Tracked in plan Discoveries as follow-ups **A7b** and **A7c**.

### Implementation shape

Approach (1) from the plan: an explicit `order: list(key)` and
`dead: MapSet.t()` on `Lua.VM.Table`, with a single `Table.put/3` write
surface that maintains both. `Table.next_entry/2` walks `order`, skips
dead-key slots, and returns `:invalid_key` (sentinel) when the key was
never seen — `lua_next` lifts that into the user-facing
`bad argument #2 to 'next' (invalid key to 'next')` error.

Routed through `Table.put/3`:

- `lib/lua/vm/state.ex` — `set_global/3`, `alloc_table/2` (via
  `Table.from_data/1` to seed `order` from initial map keys).
- `lib/lua/vm/executor.ex` — `set_list` (both forms), `table_newindex`.
- `lib/lua/vm/stdlib.ex` — `rawset`.
- `lib/lua/vm/stdlib/table.ex` — `table_insert`, `table_remove`,
  `table_sort` (via `Table.replace_data/2`), `table_move`.
- `lib/lua.ex` — public-API `set_in_table` helpers.

`Table.put_data/3` is kept for callers that only have a raw data map.

### Bonus: `table.remove` shape correction

While converting `table_remove`, lined it up with Lua 5.3 ltablib.c's
`tremove`:

- `table.remove(t)` when `#t == 0` removes (and returns) `t[0]` instead
  of returning `nil` early. nextvar.lua line 362 relies on this.
- `table.remove(t, pos)` raises `position out of bounds` when `pos < 1`
  or `pos > #t + 1` (with `pos == #t` shortcut so `table.remove(t, 0)`
  works when `#t == 0`). Previously returned `nil` silently. nextvar.lua
  line 382 relies on the raise.

### Changes

```
 .agents/plans/A7a-nextvar-dead-keys.md | 105 +++++++++++++++++++--
 lib/lua.ex                             |   8 +-
 lib/lua/vm/executor.ex                 |  47 ++--------
 lib/lua/vm/state.ex                    |  10 +-
 lib/lua/vm/stdlib.ex                   |  56 +++--------
 lib/lua/vm/stdlib/table.ex             |  90 ++++++++----------
 lib/lua/vm/table.ex                    | 161 ++++++++++++++++++++++++++++--
 test/lua/vm/nextvar_dead_keys_test.exs | 153 +++++++++++++++++++++++++++++
 8 files changed, 478 insertions(+), 152 deletions(-)
```

### Verification

```
mix format
mix test                              # 1430 tests, 0 failures, 31 skipped
mix test --only lua53                  # 5 ready / 24 skipped (unchanged)
```

- Table representation rewrite, faster `next`/`pairs`/`ipairs`, stricter
  NaN-key handling.
- Promoting other suite files to `@ready_tests`.
- Promoting `nextvar.lua` itself — see Discoveries in the plan; blocked
  by pre-existing gaps tracked as A7b (table-library metamethods) and
  A7c (numeric-`for` string coercion).

### DIFF
--- a/lib/lua/vm/stdlib/table.ex
+++ b/lib/lua/vm/stdlib/table.ex
@@ -21,6 +21,7 @@ defmodule Lua.VM.Stdlib.Table do
   alias Lua.VM.ArgumentError
   alias Lua.VM.State
   alias Lua.VM.Stdlib.Util
+  alias Lua.VM.Table
 
   @impl true
   def lib_name, do: "table"
@@ -46,8 +47,7 @@ defmodule Lua.VM.Stdlib.Table do
     # Insert at end
     table = State.get_table(state, tref)
     len = get_table_length(table)
-    new_data = Map.put(table.data, len + 1, value)
-    state = State.update_table(state, tref, fn _ -> %{table | data: new_data} end)
+    state = State.update_table(state, tref, fn t -> Table.put(t, len + 1, value) end)
     {[], state}
   end
 
@@ -62,18 +62,21 @@ defmodule Lua.VM.Stdlib.Table do
         details: "position out of bounds"
     end
 
-    # Shift elements from pos to len one position up
-    new_data =
-      len..pos
-      |> Enum.reduce(table.data, fn i, acc ->
-        case Map.get(acc, i) do
-          nil -> acc
-          val -> Map.put(acc, i + 1, val)
-        end
-      end)
-      |> Map.put(pos, value)
+    # Shift elements from pos..len up by one (write top-down so we don't
+    # clobber unread slots), then drop the new value into `pos`.
+    state =
+      State.update_table(state, tref, fn t ->
+        t =
+          Enum.reduce(len..pos//-1, t, fn i, acc ->
+            case Map.get(acc.data, i) do
+              nil -> acc
+              val -> Table.put(acc, i + 1, val)
+            end
+          end)
 
-    state = State.update_table(state, tref, fn _ -> %{table | data: new_data} end)
+        Table.put(t, pos, value)
+      end)
+
     {[], state}
   end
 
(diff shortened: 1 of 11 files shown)


test/lua/vm/nextvar_dead_keys_test.exs
```diff
@@ -0,0 +1,154 @@
+defmodule Lua.VM.NextvarDeadKeysTest do
+  @moduledoc """
+  Regressions for dead-key iteration semantics surfaced by `nextvar.lua`
+  (plan A7a, follow-up to A7).
```
davydog187: stop referring to plans. These are ephemeral resources

lib/lua/vm/stdlib/table.ex
```diff
@@ -62,18 +62,21 @@ defmodule Lua.VM.Stdlib.Table do
...
-        case Map.get(acc, i) do
-          nil -> acc
-          val -> Map.put(acc, i + 1, val)
-        end
+    # Shift elements from pos..len up by one (write top-down so we don't
+    # clobber unread slots), then drop the new value into `pos`.
+    state =
+      State.update_table(state, tref, fn t ->
+        t =
+          Enum.reduce(len..pos//-1, t, fn i, acc ->
```
davydog187, line 70: do we have tests to cover the edge cases here?

lib/lua/vm/stdlib/table.ex
```diff
@@ -143,6 +136,34 @@ defmodule Lua.VM.Stdlib.Table do
...
+        # Mirrors the loop in Lua 5.3 ltablib.c tremove: after the shift,
+        # `pos` is incremented to `size`, and the cleared slot is t[pos].
+        # When pos == 0 (and len == 0) we clear t[0]. When pos == len we
+        # clear t[len]. When pos == len + 1 (a no-op remove) we clear
+        # t[len + 1], which was already nil.
+        clear_at = max(pos, len)
+
+        t =
+          if pos < len do
+            Enum.reduce((pos + 1)..len//1, t, fn i, acc ->
```
davydog187, line 154: same, do we have tests that cover these edge cases? (ideally property tests but example tests are fine)

test/lua/vm/nextvar_dead_keys_test.exs
davydog187: Fixed in a5ad71b — moduledoc now describes the §6.1 contract directly, no plan references.

lib/lua/vm/stdlib/table.ex
davydog187: Added 7 example tests in a5ad71b under `describe "table.insert edge cases"` covering: insert at pos 1 (full shift up), insert at `#t + 1` (no shift), insert into empty/single-element tables, hash-key preservation across a shift, and out-of-bounds positions raising. Theres also a round-trip property at the bottom of the file that exercises this loop against `table.remove`.

lib/lua/vm/stdlib/table.ex
davydog187: Added 8 example tests in a5ad71b under `describe "table.remove edge cases"` covering: remove at pos 1 (full shift down), remove at `#t` (no shift), `pos == #t + 1` returns nil without mutation, the Lua 5.3 `pos == 0` special case (returns `t[0]` when `#t == 0`, raises when `#t > 0`), `pos < 0` raising, single-element list, empty table, and hash-key preservation across a shift. Plus a property test below the example block: `insert(t, pos, v)` followed by `remove(t, pos)` is the identity for the sequence portion — exercises both shift loops together.

PR #214, opened 2026-05-08
feat(error): every runtime error carries line and source info

davydog187:
## Threaded line/source info reaches every runtime error message

Plan: [`.agents/plans/A18-error-line-source-info.md`](https://github.com/tv-labs/lua/blob/fix/error-line-source-info/.agents/plans/A18-error-line-source-info.md)

### Goal

Every Lua runtime error a user sees should include the source file and
line number where the offending operation lives. Today the executor
threads `line` through every CPS dispatch, the compiler emits
`{:source_line, _, _}` markers, and the exception structs carry
`:line` / `:source` / `:call_stack` fields — but most `raise` sites in
the VM omit those fields, and the public `Lua.RuntimeException` wrapper
re-raises with only the formatted message string, dropping the
structured fields entirely.

### Before

```elixir
iex> Lua.eval!(Lua.new(), "local z = nil\nz()")
** (Lua.RuntimeException) Lua runtime error: ... attempt to call a nil value
   # message has no line, e.line is nil, e.source is nil
```

### After

```elixir
iex> Lua.eval!(Lua.new(), "local z = nil\nz()", source: "demo.lua")
** (Lua.RuntimeException) Lua runtime error: ... at demo.lua:1: attempt to call a nil value
   # e.line == 1, e.source == "demo.lua"
```

### Success criteria

- [x] `mix test` passes (1577 tests, 51 properties, 52 doctests, 0 failures — count went up by 7 with the new tests).
- [x] `mix test test/lua/error_messages_test.exs` passes (18/18).
- [x] New tests: every arithmetic / concat / index / call / assert TypeError or AssertionError raised during execution has non-nil `:line` and matching `:source`.
- [x] New test: `Lua.eval!` with a `source:` opt threads that name to `proto.source` so errors say `at script.lua:N:`.
- [x] `mix test --only lua53` still 5/29 (no regression).
- [x] Manual smoke: `Lua.eval!(Lua.new(), "local z = nil\nz()", source: "demo.lua")` produces a message containing `at demo.lua:1:`.

### Changes

```
 .agents/plans/A18-error-line-source-info.md       | 212 ++++++++++++++++++++
 .agents/plans/A19-error-line-info-native-funcs.md |  90 +++++++++
 lib/lua.ex                                        |  33 ++--
 lib/lua/runtime_exception.ex                      |  21 +-
 lib/lua/vm/executor.ex                            | 231 +++++++++++++++++-----
 test/lua/error_messages_test.exs                  | 104 ++++++++++
 test/support/lua_test_case.ex                     |   6 +-
 7 files changed, 622 insertions(+), 75 deletions(-)
```

Implementation:

- `Lua.RuntimeException` gains `:line`, `:source`, `:call_stack` fields and copies them off VM exceptions in the catchall `exception/1` clause.
- `Lua.eval!` accepts a `source:` option (default `"<eval>"`) and forwards it to `Lua.Compiler.compile/2`.
- The eval rescues now pass the original VM exception (not just its message) to `Lua.RuntimeException` so structured fields survive.
- Inside the executor, a private `with_context/4` wrapper catches `TypeError`/`RuntimeError`/`AssertionError` from helper calls and re-raises them with `:line` / `:source` / `:call_stack` filled in from the surrounding dispatch. Applied to arithmetic, bitwise, concat, compare, length, negate, table indexing, and native-function call dispatch.
- `test/support/lua_test_case.ex` passes the suite filename as `source:` so triage gets `at pm.lua:7:` instead of `<eval>:7:`.

### Discoveries

Implementation chose strategy (b) from the plan — a single executor-level wrapper rather than per-helper signature changes. TCO on `do_execute/8` is preserved (the wrap only guards helper calls, not the outer recursion).

Out-of-scope items surfaced during implementation:

- Compiler `source_line` emission has off-by-ones in some cases. New smoke tests (`local x = 1\nlocal s = "hello"\nprint(s * x)`) report line 2 when the operation is on line 3. The line-tracking infrastructure works; the compiler emits its `source_line` markers a beat early. Logged for follow-up — not blocking, since "non-nil line" is what A18 commits to.
- Stdlib raise sites (`string.upper(nil)`, `table.insert` bad arg, etc.) still raise from helper paths that bypass the executor wrap. A19 (also added in this PR as a `status: blocked` plan) covers those.

### Verification

```bash
mix format
mix test                  # 1577 / 0 failures
mix test --only lua53     # 5/29, unchanged
```

- `assert()`, `error()`, and other stdlib raises whose helpers don't have `line` in scope. Covered by A19 (drafted, blocked on this).
- Improving `call_stack` content beyond what's already plumbed.
- Changing the formatted error layout. We only ensure the data is populated.
- Source-mapping past macro expansion or `load()` chunks.

### DIFF
--- a/.agents/plans/A18-error-line-source-info.md
+++ b/.agents/plans/A18-error-line-source-info.md
@@ -1,0 +1,212 @@
+---
+id: A18
+title: Threaded line/source info reaches every runtime error message
+issue: null
+pr: null
+branch: fix/error-line-source-info
+base: main
+status: in-progress
+direction: A
+unlocks:
+  - faster suite triage (failures point at a specific line)
+  - meets the "great error messages" promise of the library
+---
+
+## Goal
+
+Every Lua runtime error a user sees should include the source file and
+line number where the offending operation lives. Today the executor
+threads `line` through every CPS dispatch, the compiler emits
+`{:source_line, _, _}` markers, and the exception structs carry
+`:line` / `:source` / `:call_stack` fields — but most `raise` sites in
+the VM omit those fields, and the public `Lua.RuntimeException` wrapper
+re-raises with only the formatted message string, dropping the
+structured fields entirely.
+
+This plan closes the gap for every `raise` site in the executor where
+`line` is in scope, fixes the wrapper to preserve structured fields,
+and threads `source:` through `Lua.eval!`. Native-function raise sites
+(`assert`, `error`, stdlib type checks) are deferred to A19 — they need
+a separate mechanism since native callees don't currently receive
+line.
+
+## Out of scope
+
+- `assert()`, `error()`, and other stdlib raises that have no `line` in
+  scope. These need an architectural change (push the calling line into
+  `state` before invoking native funcs, or thread it as an argument).
+  That goes in a follow-up plan A19.
+- Improving `call_stack` content (we just propagate what's already
+  there).
+- Changing the formatted error layout. We only ensure the data is
+  populated; the formatter already renders `at <source>:<line>:` when
+  given non-nil values.
+- Source-mapping past macro expansion or `load()` chunks. We use the
+  source name the compiler was given and the line the executor is
+  currently at; nothing fancier.
+
+## Success criteria
+
+- [ ] `mix test` passes (≥ 1670, currently 1670 + 51 prop + 52 doctest).
+- [ ] `mix test test/lua/error_messages_test.exs` passes (it already does;
+      this guards against regression).
+- [ ] New test: every arithmetic/concat/index/compare TypeError raised
+      during execution has non-nil `:line` and matching `:source`.
+- [ ] New test: `Lua.eval!(Lua.new(), src)` where `src` triggers a
+      type error produces an exception whose `:line` and `:source`
+      fields are populated (i.e. the wrapper doesn't strip them).
+- [ ] New test: `Lua.eval!` with a `source:` opt threads that name to
+      `proto.source` so errors say `at script.lua:N:` instead of
+      `at -no-source-:N:`.
+- [ ] `mix test --only lua53` still passes 5/29 (we don't regress; some
+      previously-mute assertion failures may now print line info but
+      the file-level pass/fail count won't change for this plan).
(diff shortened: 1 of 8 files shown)


davydog187:
## Performance revision

Following review feedback that the initial \`with_context\` wrapper would add overhead, I benchmarked the approach and confirmed: **~9% slowdown on \`fib(30)\`** in the benchee harness. That's not acceptable for a 1.0 library where perf is part of the contract.

Investigated alternatives:

| Approach | fib(30) slowdown | Notes |
|---|---|---|
| \`with_context\` (closure + try/rescue) per op | ~9% | Original PR — too slow |
| Process dict at every \`:source_line\` opcode | ~7% | One \`Process.put\` per Lua statement is too frequent |
| **Hybrid C: extra args + process dict at native boundary** | **~2-3%** | What I shipped |
| Inline try/rescue per opcode (no closure) | ~5% on a single opcode (compounds across all opcodes) | Less invasive but adds up |

Switched to **Hybrid C**:

1. **In-executor helpers** (\`safe_add\`, \`safe_compare_lt\`, \`concat_coerce\`, \`to_integer!\`, \`index_value\`, etc.) take \`line, source\` as args. The metamethod-dispatch closures already captured args pre-A18, so adding two more captures is essentially free on the BEAM.
2. **Process dict only at the \`:native_func\` call boundary** (\~3 sites). Stash calling line/source before invoking the callback, restore after. \`assert\`, \`error\`, and other stdlib raise sites read via \`Lua.VM.Executor.current_position/0\`.
3. **Save/restore in \`Lua.VM.Executor.execute/5\`** for re-entrancy and isolation between sequential calls.

The \`:source_line\` opcode is **not** touched — writing to process dict on every Lua statement was the dominant cost in the all-process-dict variant.

Re-entrancy is handled: nested \`Lua.eval!\` (e.g. an Elixir callback that itself calls into Lua) saves and restores the outer position via \`try/after\` in \`execute/5\`. Sequential top-level calls don't leak either.

See commit \`3e6827f\` for the implementation diff and the updated plan file for the full discoveries section.

PR #221, opened 2026-05-21
fix(error): prune internal frames from Lua.RuntimeException stack

davydog187:
## Suppress Elixir VM frames from Lua.RuntimeException stack traces

Plan: [`.agents/plans/A37-prune-elixir-stack-trace.md`](https://github.com/tv-labs/lua/blob/fix/prune-elixir-stack-trace/.agents/plans/A37-prune-elixir-stack-trace.md)

### Goal

When a Lua program raises a runtime error through `Lua.eval!`, the user sees the *Lua* stack trace and source location, not a six-frame Elixir stack trace pointing into our executor internals.

Before:

```
** (Lua.RuntimeException) Lua runtime error: Runtime Type Error
  at <eval>:2:
  attempt to perform arithmetic on a string value
    (lua 1.0.0-rc.1) lib/lua/vm/executor.ex:1810: Lua.VM.Executor.raise_arith_type_error/3
    (lua 1.0.0-rc.1) lib/lua/vm/executor.ex:1760: Lua.VM.Executor.invoke_metamethod/4
    (lua 1.0.0-rc.1) lib/lua/vm/executor.ex:859: Lua.VM.Executor.do_execute/8
    (lua 1.0.0-rc.1) lib/lua/vm/executor.ex:76: Lua.VM.Executor.execute/5
    (lua 1.0.0-rc.1) lib/lua/vm.ex:26: Lua.VM.execute/2
    (lua 1.0.0-rc.1) lib/lua.ex:459: Lua.eval!/3
    iex:6: (file)
```

After:

```
** (Lua.RuntimeException) Lua runtime error: Runtime Type Error
  at <eval>:2:
  attempt to perform arithmetic on a string value
    (lua 1.0.0-rc.1) lib/lua/vm.ex:26: Lua.VM.execute/2
    (lua 1.0.0-rc.1) lib/lua.ex:459: Lua.eval!/3
    iex:6: (file)
```

### Success criteria

- [x] `mix test` passes (1672 → 1681, +9 tests, 0 failures)
- [x] Executor / compiler / parser / lexer submodule frames pruned from rescued runtime errors
- [x] User-facing frames (`Lua.eval!`, the user's call site) preserved
- [x] `LUA_DEBUG_STACK=1` env var restores full frames for debugging
- [x] `Lua.VM.InternalError` keeps full stack — those are library bugs, not Lua errors
- [x] A user app module under a `Lua.*` namespace is never pruned (verified with unit test)
- [x] `mix test --only lua53` unchanged

### Changes

```
 .agents/plans/A26-error-message-quality.md    |  31 +++-
 .agents/plans/A36-parser-error-positions.md   | 252 +++++++++++++++++++++
 .agents/plans/A37-prune-elixir-stack-trace.md | 199 ++++++++++++++++
 lib/lua.ex                                    |  64 ++++++-
 test/lua/runtime_exception_test.exs           | 179 ++++++++++++++++
 5 files changed, 720 insertions(+), 5 deletions(-)
```

### Verification

```bash
mix format
mix compile --warnings-as-errors  # clean
mix test                          # 1681 passing, 0 failing
mix test --only lua53             # 6 passing, 23 skipped (unchanged)
```

Manual smoke through `mix run`:

```elixir
Lua.eval!(~S\"\"\"
function foo(thing)
  return thing * 2
end
return foo("yoo")
\"\"\")
# Stack contains: Lua.VM.execute/2, Lua.eval!/3, user frame
# NO Lua.VM.Executor.* frames

# With LUA_DEBUG_STACK=1:
# Stack contains the full chain including
# Lua.VM.Executor.raise_arith_type_error/3,
# Lua.VM.Executor.invoke_metamethod/4, etc.
```

### Discoveries

- `Lua.VM.execute/2` (the top-level VM entrypoint module, NOT a submodule like `Lua.VM.Executor`) is preserved by the prune, because the prefix match `"Elixir.Lua.VM."` requires a trailing dot. This is intentional — `Lua.VM` is part of the public boundary and dropping it would leave a visible gap between `Lua.eval!` and the user's frame.
- Two rescue blocks needed updating: one for the `binary` script variant of `eval!/3` and one for the `Lua.Chunk` variant. Both get identical treatment.
- `Lua.VM.InternalError` is rescued before the generic `RuntimeError / TypeError / AssertionError` branch and re-raised with the original `__STACKTRACE__` so library debugging still has full visibility.
- Frame-shape handling: the helper pattern-matches both 4-tuple and 5-tuple stacktrace entries and falls through to "not internal" for any other shape, so we never crash on an unexpected frame format.

- Touching the Lua-level call stack rendering (`:call_stack` field, rendered by `Lua.VM.ErrorFormatter`) — that stays as-is; A26 polishes it.
- Hiding `Lua.InternalError` frames — those are bugs in us, kept fully visible.
- Pruning frames from `Lua.CompilerException` — compile errors raise before any VM execution; A36 covers the parser side via position info, not via stack pruning.
- A `:debug_stack` option on `Lua.eval!` — env var only, matching how Elixir itself toggles internal-stack visibility.

### Related

This is the first of three plans tightening up error rendering:

- **A37 (this PR)** — Elixir stack prune. Tiny, high-impact, establishes a clean visual baseline.
- **A36** — Parser errors carry position; rewrite the "Expression statement must be a function call" message.
- **A26** — Error gallery + suggestion polish, now unblocked.

### DIFF
--- a/lib/lua.ex
+++ b/lib/lua.ex
@@ -23,7 +23,7 @@ defmodule Lua do
   # for future-proofing. Suppress dialyzer warnings on those defensive branches.
   @dialyzer {:no_match, eval!: 2, eval!: 3, parse_chunk: 1}
 
-  defstruct [:state]
+  defstruct [:state, debug: false]
 
   @default_sandbox [
     [:io, :stdin],
@@ -83,17 +83,20 @@ defmodule Lua do
   ## Options
   * `:sandboxed` - list of paths to be sandboxed, e.g. `sandboxed: [[:require], [:os, :exit]]`
   * `:exclude` - list of paths to exclude from the sandbox, e.g. `exclude: [[:require], [:package]]`
+  * `:debug` - (default `false`) when `true`, internal Lua VM frames are preserved in stack traces
+    instead of being pruned. Useful when debugging library bugs.
   """
   def new(opts \\ []) do
-    opts = Keyword.validate!(opts, sandboxed: @default_sandbox, exclude: [])
+    opts = Keyword.validate!(opts, sandboxed: @default_sandbox, exclude: [], debug: false)
     exclude = Keyword.fetch!(opts, :exclude)
+    debug = Keyword.fetch!(opts, :debug)
 
     state = Lua.VM.Stdlib.install(State.new())
 
     opts
     |> Keyword.fetch!(:sandboxed)
     |> Enum.reject(fn path -> path in exclude end)
-    |> Enum.reduce(%__MODULE__{state: state}, &sandbox(&2, &1))
+    |> Enum.reduce(%__MODULE__{state: state, debug: debug}, &sandbox(&2, &1))
   end
 
   @doc """
@@ -490,10 +493,10 @@ defmodule Lua do
     # `Lua.RuntimeException.exception/1` clause for arbitrary exceptions
     # picks up `:line`, `:source`, and `:call_stack`.
     e in [RuntimeError, TypeError, AssertionError] ->
-      reraise Lua.RuntimeException, e, prune_lua_internals(__STACKTRACE__)
+      reraise Lua.RuntimeException, e, prune_lua_internals(__STACKTRACE__, lua.debug)
 
     e ->
-      reraise Lua.RuntimeException, e, prune_lua_internals(__STACKTRACE__)
+      reraise Lua.RuntimeException, e, prune_lua_internals(__STACKTRACE__, lua.debug)
   end
 
   def eval!(%__MODULE__{state: state} = lua, %Lua.Chunk{prototype: proto}, opts) do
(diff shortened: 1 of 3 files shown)


lib/lua.ex
```diff
@@ -509,14 +515,64 @@ defmodule Lua do
...
+
+  # Removes internal executor / compiler / parser / lexer frames from a
+  # rescued stacktrace so the user sees only frames they can act on:
+  # their own call site and the public `Lua.eval!` boundary.
+  #
+  # Set `LUA_DEBUG_STACK=1` in the environment to disable pruning when
+  # debugging library bugs.
+  @doc false
+  def prune_lua_internals(stacktrace) do
+    if debug_stack?() do
```
davydog187: lets store this in the lua struct itself rather than relying on env vars. Let the user pass `debug: true`. by default it will be false

lib/lua.ex
```diff
@@ -509,14 +515,64 @@ defmodule Lua do
...
+  ]
+
+  # Removes internal executor / compiler / parser / lexer frames from a
+  # rescued stacktrace so the user sees only frames they can act on:
+  # their own call site and the public `Lua.eval!` boundary.
+  #
+  # Set `LUA_DEBUG_STACK=1` in the environment to disable pruning when
+  # debugging library bugs.
+  @doc false
+  def prune_lua_internals(stacktrace) do
```
davydog187: this should be private

test/lua/runtime_exception_test.exs
```diff
@@ -431,4 +433,181 @@ defmodule Lua.RuntimeExceptionTest do
...
+      ]
+
+      pruned = Lua.prune_lua_internals(stack)
+      modules = stacktrace_modules(pruned)
+
+      assert Something in modules
+      refute Executor in modules
+      assert Lua in modules
+    end
+  end
```
davydog187: instead of testing directly, lets cover these through integration tests so we can leave the `prune_lua_internals` function private

PR #228, opened 2026-05-21
feat(error): name the callee/target in attempt to call/index errors

davydog187:
## Summary

- Annotates `:call`, `:get_field`, `:set_field`, `:get_table`, `:set_table`, and `:self` instructions with a compile-time `name_hint` derived from the AST (`{:global, name}` / `{:local, name}` / `{:upvalue, name}` / `{:field, name}` / `{:method, name}`).
- Executor formats the hint into the runtime error message, mirroring PUC-Lua.
- Fixes `:set_field` / `:set_table` to raise `:index_non_table` instead of `MatchError` when the target isn't a table.

### Before
```
> Lua.eval!("return foo()")
attempt to call a nil value

> Lua.eval!("foo.bar()")
attempt to index a nil value
```

### After
```
> Lua.eval!("return foo()")
attempt to call a nil value (global 'foo')

> Lua.eval!("foo.bar()")
attempt to index a nil value (global 'foo')

> Lua.eval!("local t = {}; t.bar()")
attempt to call a nil value (field 'bar')

> Lua.eval!("local t = {}; t:baz()")
attempt to call a nil value (method 'baz')

> Lua.eval!("local n = 5; return n.x")
attempt to index a number value (local 'n')
```

PUC-Lua recovers the same names via debug info threaded through bytecode; doing it at compile time is simpler here because the AST is still in scope when we emit the instruction.

## Test plan

- [x] `mix test test/lua/error_messages_test.exs` — 36 tests including 10 new cases covering global/local/upvalue/field/method calls, set-side index, non-function/non-table targets, and the anonymous-callee (no-hint) case.
- [x] `mix test` — full suite (1778 tests, 30 skipped) passes.
- [x] Manual `iex -S mix` smoke test for each error path.

### DIFF
--- a/test/lua/error_messages_test.exs
+++ b/test/lua/error_messages_test.exs
@@ -3,10 +3,15 @@ defmodule Lua.ErrorMessagesTest do
 
   alias Lua.Compiler
   alias Lua.Parser
+  alias Lua.RuntimeException
   alias Lua.VM
   alias Lua.VM.ArgumentError
   alias Lua.VM.State
   alias Lua.VM.TypeError
+
+  defp error_message(code) do
+    Exception.message(assert_raise(RuntimeException, fn -> Lua.eval!(code) end))
+  end
 
   describe "beautiful error messages" do
     test "calling nil value shows helpful error" do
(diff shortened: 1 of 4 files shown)


test/lua/error_messages_test.exs
```diff
@@ -92,6 +97,65 @@ defmodule Lua.ErrorMessagesTest do
...
+      end
+      return inner()
+      """
+
+      assert error_message(code) =~ "attempt to call a nil value (upvalue 'x')"
+    end
+
+    test "call-nil error names a field" do
+      assert error_message("local t = {}\nt.bar()") =~
+               "attempt to call a nil value (field 'bar')"
```
davydog187: For the case where we're invoking functions and methods, it would be nice it if provided this information, i.e. I want to see that it was the `t.bar()` that failed (the 't' symbol is not mentioned in the current). Its definitely important to note which is nil as well (bar is nil, not t)

PR #235, opened 2026-05-22
perf(vm): compile Lua prototypes to BEAM modules

davydog187:
## Plan: B5a — Erlang codegen foundation

Plan: [`.agents/plans/B5a-erlang-codegen-foundation.md`](.agents/plans/B5a-erlang-codegen-foundation.md)
Parent strategic plan: [`.agents/plans/B5-compile-prototypes-to-erlang.md`](.agents/plans/B5-compile-prototypes-to-erlang.md)

### Goal

Land the foundation for compiling Lua prototypes to BEAM modules
via `:compile.forms/2`. A compiled prototype's call goes through a
new `{:compiled_closure, mod, fun, upvalues, proto}` value type,
bypassing the interpreter's register-tuple construction and per-opcode
dispatch loop entirely. This first PR covers arithmetic, comparison,
logical ops, conditional `:test`, single-result `:call`,
single-value `:return`, and the common `_ENV.name` lookup path.

### Scope

Supported in this PR:

- Constants and moves: `:load_constant`, `:load_boolean`, `:load_nil`,
  `:move`, `:source_line`, `:scope`
- Upvalues + globals: `:get_upvalue`, `:get_open_upvalue`,
  `:load_env`, `:get_global`
- `_ENV.name` field access: `:get_field` with binary literal name
  (inlines the no-metatable fast path; metatable case delegates to
  `Executor.index_value/6`)
- Arithmetic with integer fast path: `:add`, `:subtract`, `:multiply`;
  slow-path-only for `:divide`, `:floor_divide`, `:modulo`, `:power`,
  `:negate`
- Comparison with number fast path: `:less_than`, `:less_equal`,
  `:greater_than`, `:greater_equal`; slow-path-only for `:equal`,
  `:not_equal`
- Logical `:not`
- Conditional `:test` and `:test_true` — restricted to branches that
  terminate via `:return` (no SSA-merging in B5a)
- `:call` with single-result returns; routes through
  `call_function_with_position` which bridges native-callback position
  tracking but no-ops for pure Lua-to-Lua calls.

Out of scope (deliberately falling back to interpreter):

- Tables → B5c
- Closures, varargs, multi-return → B5d
- Error position fidelity inside compiled raises → B5e
- `:goto`/`:label`, loops (`:numeric_for`, `:while_loop`, etc.)

All-or-nothing per prototype: a prototype containing any unsupported
opcode falls back to interpretation in its entirety. Sub-prototypes
compile independently.

### Success criteria

- [x] `Lua.Compiler.Erlang.compile/1` exists and returns
      `{:ok, proto_with_compiled_module_set}` for covered prototypes
- [x] `Lua.VM.CompiledModule` value type wired through
      `Executor.call_function/3` and the `:call` opcode dispatch
- [x] Every covered opcode lowered in `Lua.Compiler.Erlang.Opcodes`
- [x] Uncovered opcodes trigger fallback — never crash
- [x] Closure construction (`:closure`) emits `:compiled_closure`
      when the nested prototype compiled, else `:lua_closure`
- [x] `mix test`: 1705 tests + 51 properties + 55 doctests, **0
      failures**
- [x] `mix test --only lua53`: 29 tests, **0 failures**
- [ ] fib(25) beats Luerl by ≥5x — **not met** (achieves ~1.1x).
      The throw/catch overhead on non-tail returns and the
      register-tuple `setelement/3` churn dominate; B5b/B5c/B5d will
      close the gap as more opcodes inline.
- [x] No workload regresses

### Perf

`fib(30)`, full mode:

| Implementation | Mean | vs main | vs Luerl |
|---|---|---|---|
| main           | ~970 ms | 1.00x   | 0.74x slower |
| B5a (this PR)  | ~670 ms | **1.45x faster** | **1.07x faster** |
| Luerl          | ~720 ms | 1.35x faster | baseline |
| C Lua (luaport)| ~27 ms  | 36x faster | 27x faster |

The compiled path beats Luerl modestly today. The 5x stretch target
is held back primarily by:

- **`throw/catch` for non-tail returns** (~8% of CPU). This PR
  optimises the function-tail `:return` to natural-return; returns
  inside `:test` branches still throw. B5e (error fidelity) will
  revisit the throw/catch shape.
- **`setelement/3` per opcode write** (~22% of CPU). Equivalent to
  the interpreter's register-tuple cost. Register promotion to SSA
  Erlang variables (deferred follow-up) eliminates this.
- **`apply_arith_op` / `index_value` calls when the inline fast path
  doesn't fire.** B5c adds table-opcode coverage which inlines more
  paths.

### Changes

- `lib/lua/compiler/erlang.ex` — top-level compile/load orchestration
- `lib/lua/compiler/erlang/codegen.ex` — abstract-forms generation
- `lib/lua/compiler/erlang/opcodes.ex` — per-opcode lowering
- `lib/lua/compiler/erlang/runtime.ex` — generated-code runtime helpers
- `lib/lua/compiler/prototype.ex` — `compiled_module` field
- `lib/lua/compiler.ex` — wire codegen into `Lua.Compiler.compile/2`
- `lib/lua/vm.ex` — top-level execute dispatches to compiled module
- `lib/lua/vm/executor.ex` — `:compiled_closure` clauses in
  `call_function/3` and the `:call` opcode; `apply_arith_op/6`,
  `apply_unary_op/5`, `apply_compare_op/6`,
  `call_function_with_position/5` public helpers; `index_value/6`
  promoted to public
- `lib/lua/vm/value.ex`, `lib/lua/util.ex`, `lib/lua/api.ex`,
  `lib/lua/vm/display.ex`, `lib/lua/vm/stdlib*.ex`, `lib/lua.ex` —
  add `:compiled_closure` clauses everywhere `:lua_closure` was
  pattern-matched

### Verification

```
mix format
mix test                       # 1705 tests, 0 failures
mix test --only lua53          # 29 tests, 0 failures
MIX_ENV=benchmark mix run benchmarks/fibonacci.exs
```

### Known limitations (followed up in B5b–B5e)

- Every prototype gets a fresh module name; loaded modules persist
  until BEAM exit. **B5b** introduces the content-addressable
  ref-counted cache.
- `:get_field` with non-binary name, all other table opcodes, and
  closures fall back. **B5c** and **B5d** cover them.
- Errors raised from compiled code carry the codegen-time `:source_line`
  but not full position fidelity. **B5e** adds try/catch with
  `pc_to_line` tables.
- One observed `:erl_lint :unsafe_var` warning logs (not a failure)
  for prototypes with a specific shape involving register write
  inside `:test` branches that then continue. The prototype safely
  falls back in that case.


davydog187:
Closing unmerged.

The compile-to-BEAM-modules approach is fundamentally unsafe for an embedded Lua library:

- Every successful compile mints a fresh module atom (`lua_proto_<n>`) and loads a BEAM binary. Atoms are never GC'd. Default cap ~1M atoms, then the VM aborts.
- Purging is not a viable mitigation. `:code.soft_purge/1` is blocked by any process executing in the module — including Lua coroutines suspended mid-call — and `:code.purge/1` kills those processes. A host has no safe time to reclaim modules referenced by in-flight execution.
- Lua's own `load(untrusted_input)` builtin makes this an atom-table-exhaustion DoS reachable from any host that exposes runtime Lua compilation. That's the whole point of an embedded Lua.

The planned B5b (content-addressable cache + ref counting) does not fix this. Ref counting can't tell whether a process is *currently executing* the code; soft-purge can refuse and hard-purge can crash. Content addressing only bounds the leak if the set of distinct Lua programs is bounded, which a `load(timestamp)` loop defeats.

Luerl, the reference Erlang Lua implementation, never calls `compile:*` or `code:*`. Its compiled chunks are plain Erlang data (tuples and records) walked by a hand-written interpreter (`luerl_emul:emul_1/7`). GC reclaims them like any other term.

Direction B is pivoting to the same constraint: emit a dense register-based **bytecode** as plain Erlang data, and ship a single hand-written **dispatcher module** in `lib/` that interprets it. No runtime BEAM module generation. No atoms minted per compile. No lifecycle to manage.

The codegen *structure* in this PR survives the pivot — the per-opcode lowering decisions in `lib/lua/compiler/erlang/opcodes.ex` (integer fast-path inlining, comparison shape, branch encoding) port to the bytecode encoder. The Erlang-AST output target and `:compile.forms` orchestration delete entirely.

See updated plans:
- `.agents/plans/B5-dispatcher-and-bytecode.md` (replacement parent)
- `.agents/plans/B5a-v2-dispatcher-foundation.md` (replacement for this PR's scope)
- `.agents/plans/B5-compile-prototypes-to-erlang.md` (marked superseded)
- `.agents/plans/B5a-erlang-codegen-foundation.md` (marked superseded)
- `.agents/plans/B5b-module-lifecycle.md` (marked abandoned — the leak fix that doesn't fix the leak)
- `.agents/plans/B5c-table-opcodes.md`, `B5d-closures-and-varargs.md`, `B5e-error-position-fidelity.md` (marked superseded)

Issue #252, opened 2026-05-28
Normalise stdlib error message formats to match PUC-Lua


davydog187:
## Problem

Several Lua 5.3 suite files (`math.lua`, `sort.lua`, `errors.lua`, `strings.lua`, `goto.lua`) use a `checkerror(msg, f, ...)` helper that calls `pcall(f, ...)` and asserts `string.find(err, msg)` matches a specific substring. We produce the right error *category* but the wrong *phrasing*, so dozens of assertions fail.

## Concrete templates to match

- `"wrong number of arguments to 'X'"` — when too few/many args (e.g. `table.insert({}, 2, 3, 4)`). One case already fixed in #251.
- `"bad argument #N to 'X' (T expected, got T)"` — verify the exact wording across all stdlib argument-error sites.
- `"attempt to perform 'n//0'"` — integer divide-by-zero (op token included).
- `"... has no integer representation"` — float-to-int conversions (`f | 1`, `f >> 0`, bitwise on non-representable floats).
- `"field 'X'"` prefix when arithmetic/indexing fails on a global access (e.g. `math.huge << 1` → `"field 'huge'"`).

## Why it matters

This is the biggest single lever for suite progress. Likely unlocks 60–150 individual asserts across at least 4 suite files. The work is mechanical: identify each `raise` site in `lib/lua/vm/stdlib*` and `lib/lua/vm/executor.ex`, compare wording to PUC-Lua's `luaL_error`/`luaL_argerror` output, adjust.

## Suggested approach

1. Audit the existing `Lua.VM.ArgumentError` template in `lib/lua/vm/argument_error.ex` against PUC-Lua's `luaL_argerror` format.
2. Add a single helper for "wrong number of arguments to 'X'" and route variadic stdlib functions through it.
3. For arithmetic-on-globals errors, attach the field hint that `:get_field`/`:get_global` already tracks.
4. Pin each template with a unit test under `test/lua/vm/` so the wording is enforced.

## Acceptance

- New unit tests pinning each error template.
- Re-running the suite shows progress in `math.lua`, `sort.lua`, `errors.lua`, `strings.lua` (some still deferred for other reasons, but their `checkerror` ranges should shrink or pass).

PR #258, opened 2026-05-28
feat(stdlib): utf8 library + align integer-arith error wording

### LINKED ISSUE #253: Implement utf8 standard library
## Problem

The `utf8` standard library is not implemented. `utf8.lua` from the official suite is deferred entirely on `local utf8 = require'utf8'` (yields `true` from our require cache rather than the library table).

## Scope

Lua 5.3 §6.5 — six surface functions plus one field:

- `utf8.char(...)` — codepoints → UTF-8 string
- `utf8.codepoint(s [, i [, j]])` — UTF-8 string → codepoints
- `utf8.codes(s)` — stateless iterator over `(byte_pos, codepoint)` pairs
- `utf8.len(s [, i [, j]])` — number of codepoints (or `nil, byte_pos` on invalid)
- `utf8.offset(s, n [, i])` — byte position of the *n*-th codepoint
- `utf8.charpattern` — string constant: pattern matching one valid UTF-8 byte sequence

All operations are over bytes; Elixir's `String.next_codepoint/1` handles validation for us.

## Suggested approach

1. New module `lib/lua/vm/stdlib/utf8.ex` implementing `Lua.VM.Stdlib.Library`. Follow the layout of `lib/lua/vm/stdlib/math.ex`.
2. Register via `install_library/2` in `lib/lua/vm/stdlib.ex` so `require"utf8"` resolves it from `package.loaded`.
3. New test file `test/lua/vm/utf8_test.exs` covering the well-defined cases plus invalid-sequence handling.

## Acceptance

- `mix test` passes.
- `utf8.lua` either passes outright or progresses to a well-defined later failure that we can triage separately. Remove its `:all` entry from `test/lua53_skips.exs`.

davydog187:
Bundles two T1 suite-progress issues: implements the `utf8` stdlib (#253) and normalises a handful of stdlib/executor error templates to match PUC-Lua wording (#252).

## Summary

- **New `utf8` standard library** (`lib/lua/vm/stdlib/utf8.ex`): full Lua 5.3 §6.5 surface — `char`, `codepoint`, `codes`, `len`, `offset`, `charpattern`. Validates the 0x0..0x10FFFF range, rejects overlong sequences, continuation-byte starts, and codepoints above U+10FFFF. The official suite's `utf8.lua` now passes outright and is removed from `test/lua53_skips.exs`.
- **Error wording normalisation** for integer divide/modulo by zero:
  - `1 // 0` → `"attempt to perform 'n//0'"` (was `"attempt to divide by zero"`)
  - `1 % 0` → `"attempt to perform 'n%0'"` (was `"attempt to perform modulo by zero"`)
- **`ArgumentError.wrong_number_of_arguments/1`** constructor: shared helper for the PUC-Lua "wrong number of arguments to 'X'" runtime-error template. `table.insert` now routes through it; other variadic stdlib sites can adopt as they're audited.
- Each touched template is pinned with a unit test in `test/lua/vm/error_format_test.exs`.

## Out of scope (follow-ups for #252)

Threading `(field 'X')` / `(global 'X')` / `(local 'X')` target hints into arithmetic-type errors (e.g. so `math.huge << 1` surfaces `field 'huge'`) requires plumbing a hint parameter through every `raise_arith_type_error` caller — eight callsites, plus the VM dispatcher that knows the operand's lexical origin. Left for a follow-up because it's a larger refactor than the wording fixes in this PR.

The other `:all`-deferred suite files (`math.lua`, `sort.lua`, `errors.lua`, `strings.lua`) still fail at early, non-wording-related assertions and stay `:all`-skipped; further narrowing depends on the field-hint work above plus other behavioural fixes.

## Test plan

- [x] `mix test` — 1953 passed (was 1914 on baseline), 25 skipped (was 26 — `utf8.lua` promoted out)
- [x] `mix format --check-formatted`
- [x] `mix dialyzer` — no new errors (one pre-existing on `main` in `tasks/suite_runner.ex`)
- [x] `test/lua/vm/stdlib/utf8_test.exs` — 29 tests covering happy path + every error template
- [x] `test/lua/vm/error_format_test.exs` — 9 tests pinning every PUC-aligned template

Closes #253. Progresses #252.

### DIFF
--- a/lib/lua/vm/stdlib/utf8.ex
+++ b/lib/lua/vm/stdlib/utf8.ex
@@ -63,23 +63,17 @@ defmodule Lua.VM.Stdlib.Utf8 do
     {[IO.iodata_to_binary(iolist)], state}
   end
 
-  defp encode_one(cp, _idx) when is_integer(cp) and cp >= 0 and cp <= @max_codepoint do
-    encode_codepoint(cp)
-  end
+  defp encode_one(cp, idx) do
+    cp = check_integer(cp, "utf8.char", idx)
 
-  defp encode_one(cp, idx) when is_integer(cp) do
-    raise ArgumentError,
-      function_name: "utf8.char",
-      arg_num: idx,
-      details: "value out of range"
-  end
-
-  defp encode_one(v, idx) do
-    raise ArgumentError,
-      function_name: "utf8.char",
-      arg_num: idx,
-      expected: "number",
-      got: Util.typeof(v)
+    if cp >= 0 and cp <= @max_codepoint do
+      encode_codepoint(cp)
+    else
+      raise ArgumentError,
+        function_name: "utf8.char",
+        arg_num: idx,
+        details: "value out of range"
+    end
   end
 
   defp encode_codepoint(cp) when cp < 0x80, do: <<cp>>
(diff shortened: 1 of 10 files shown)


lib/lua/vm/stdlib/utf8.ex
```diff
@@ -0,0 +1,435 @@
...
+  # ---------------------------------------------------------------------------
+  # utf8.codepoint(s [, i [, j]])
+  # ---------------------------------------------------------------------------
+
+  defp utf8_codepoint([], _state) do
+    raise ArgumentError.value_expected("utf8.codepoint", 1)
+  end
+
+  defp utf8_codepoint([s | rest], state) do
+    if !is_binary(s) do
```
davydog187:
```suggestion
    if not is_binary(s) do
```

PR #287, opened 2026-05-31
test(suite): narrow gc.lua skip to the collectgarbage no-op region

### LINKED ISSUE #260: Triage cluster — VM-level errors in gc.lua and attrib.lua
Tracked by plan: [`.agents/plans/A22-triage-vm-runtime-errors.md`](https://github.com/tv-labs/lua/blob/main/.agents/plans/A22-triage-vm-runtime-errors.md)

When this work is picked up:

3. Merge closes this issue.

davydog187:
## Goal

Narrow `gc.lua`'s whole-file skip to the region that depends on a real garbage collector, so the file's early allocation smoke tests run and report conformance. Triage sub-plan of cluster A22 (issue #260).

## Success criteria

- [x] `gc.lua` is no longer a whole-file `:all` skip; it runs with a narrowed range and passes (`456 lines skipped, 1 ranges`).
- [x] A regression test pins the `collectgarbage` stub behaviour and documents the PUC-Lua divergences: `test/lua/vm/stdlib/collectgarbage_test.exs`.
- [x] `mix test` passes (2044 passed, was 2037).
- [x] `mix test test/lua53_suite_test.exs --only lua53` passes (15 passed / 14 skipped, was 14 / 15).

## Changes

- `test/lua53_skips.exs`: narrow `gc.lua` from `lines: :all` to `lines: 167..622`, category `:unimplemented`, with a precise reason and `issue: 260`.
- `test/lua/vm/stdlib/collectgarbage_test.exs`: new regression test pinning the no-op `collectgarbage` stub and enumerating the divergences a future real collector must satisfy.
- `.agents/plans/A22a-gc-collectgarbage-stub.md`: sub-plan.

### Why

The parent cluster A22 hypothesised a leaked Elixir `MatchError` in the executor for `gc.lua`. That no longer reproduces — intervening VM work resolved it. The current first failure is a Lua-level assertion: `collectgarbage` is a no-op stub (`lib/lua/vm/stdlib.ex`) where `isrunning` always returns `true`, `count` returns `0.0`, and other modes return `0`. So `collectgarbage("stop")` cannot be honoured, and `gc.lua:168` (`assert(not collectgarbage("isrunning"))`) fails via `dosteps` at line 186.

From the `dosteps` helper (line 167) onward the file is pervasively GC-dependent: step pacing returning a boolean cycle flag, `gcinfo()` shrinking after collection, weak tables, `__gc` finalizers, threads, and emergency collections. Lines 1-165 (basic allocation) and the trailing `print('OK')` do not depend on a real collector, so they now run.

## Verification

```
mix format
mix test
# Result: 2044 passed (56 doctests, 51 properties, 1937 tests), 21 skipped
mix test test/lua53_suite_test.exs --only lua53
# Result: 15 passed, 14 skipped  (gc.lua flipped into the passing set)
```

`gc.lua` flips from `(pending initial triage)` (whole-file skip) to `(456 lines skipped, 1 ranges)` and passes.

## Out of scope

- Implementing a real garbage collector (step/stop/restart pacing, `count` shrinkage, weak tables, `__gc` finalizers). Deferred well past 1.0; the narrowed skip and the regression test track the gap.
- `attrib.lua` — permanently deferred (filesystem/package I/O).

Plan: A22a

Closes #260

### DIFF
--- a/.agents/plans/A22a-gc-collectgarbage-stub.md
+++ b/.agents/plans/A22a-gc-collectgarbage-stub.md
@@ -1,0 +1,93 @@
+---
+id: A22a
+title: "Triage gc.lua — narrow skip to the collectgarbage no-op region"
+issue: 260
+pr: null
+branch: fix/gc-vm-errors
+base: main
+status: in-progress
+direction: A
+unlocks:
+  - gc.lua
+---
+
+## Goal
+
+Narrow `gc.lua`'s whole-file skip to the region that depends on a real
+garbage collector, so the file's early allocation smoke tests run and
+report conformance.
+
+## Out of scope
+
+- Implementing a real garbage collector (step/stop/restart pacing,
+  `count` shrinkage, weak tables, `__gc` finalizers). That is a large
+  feature, deferred well past 1.0.
+- `attrib.lua` — permanently deferred (filesystem/package I/O), tracked
+  under the parent cluster.
+
+## Success criteria
+
+- [ ] `gc.lua` is no longer a whole-file `:all` skip; it runs with a
+      narrowed range and passes.
+- [ ] A regression test pins the `collectgarbage` stub behaviour and
+      documents the PUC-Lua divergences:
+      `test/lua/vm/stdlib/collectgarbage_test.exs`.
+- [ ] `mix test` passes.
+- [ ] `mix test test/lua53_suite_test.exs --only lua53` passes.
+
+## Implementation notes
+
+Parent cluster: A22 (issue #260). The parent hypothesised a leaked
+Elixir `MatchError` in the executor; that no longer reproduces —
+intervening VM work resolved it. The current first failure in `gc.lua`
+is a Lua-level assertion, not a leaked match.
+
+Root cause: `collectgarbage` is a no-op stub
+(`lib/lua/vm/stdlib.ex`). `isrunning` always returns `true`,
+`count` returns `0.0`, and every other mode returns `0`. The first
+failing assertion is `gc.lua:168` (`assert(not collectgarbage("isrunning"))`)
+reached via `dosteps(0)` at `gc.lua:186`, because the stub cannot honour
+`collectgarbage("stop")`.
+
+From the `dosteps` helper (line 167) onward the file is pervasively
+GC-dependent: step pacing returning a boolean cycle flag, `gcinfo()`
+shrinking after a collection, weak tables, `__gc` finalizers, threads,
+and emergency collections. Lines 1–165 (basic allocation) and the
+trailing `print('OK')` do not depend on a real collector.
+
+Decision: defer (triage skill §6.C). Narrow the `gc.lua` skip from
+`:all` to `167..622` with a precise reason and issue #260, and leave a
+regression test documenting the stub.
+
+The long string `[[ ... ]]` at lines 125–134 is fully inside the
+passing prefix, so the range boundary at 167 does not split it. Nothing
+declared inside 167..622 is referenced after line 622.
+
+### Files
+
+- `test/lua53_skips.exs` — narrow `gc.lua` range.
+- `test/lua/vm/stdlib/collectgarbage_test.exs` — regression test.
+
+## Verification
+
+```bash
+mix format
+mix compile --warnings-as-errors
+mix test
+mix test test/lua53_suite_test.exs --only lua53
+```
+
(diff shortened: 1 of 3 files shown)


davydog187:
## Review: PR #287 — narrow gc.lua skip to the collectgarbage no-op region

Verified the change end to end. Solid triage and a correct, conservative skip narrowing.

**Verification (re-ran locally):**
- `mix test test/lua53_suite_test.exs --only lua53` → 15 passed, 14 skipped (gc.lua flips into the passing set, matches the PR claim).
- `mix test` → 2044 passed, 21 skipped (matches).
- New `collectgarbage_test.exs` → 6 passed.

**Boundary analysis confirmed correct.** The skip range `167..622` is precise:
- The prefix (1–166) *defines* `GC1`/`GC2`/`GC` (lines 43–82) whose `repeat ... until finish` loops would spin forever without `__gc` finalizers, but `GC()` is only *called* at line 314+, fully inside the skip range. So the prefix never hits an infinite loop.
- Line 622 (`assert(collectgarbage'isrunning')`) is included in the skip even though the stub would pass it — conservative and fine. Line 624 `print('OK')` runs.
- The `[[ ... ]]` long string (125–134) is inside the passing prefix; boundary 167 doesn't split it.

**Conventions: all clean.**
- `test(suite):` scope on the implementation commit; `chore(A22a):` only on plan-only commits (allowed exception).
- `Plan: A22a` in commit body, `Closes #260` in PR + commit body.
- No Co-Authored-By trailers. No plan-id references in the test, skip reason, or any source file.
- Skip reason is precise and carries `issue: 260`.

**Minor / nits (non-blocking):**
1. `test/lua53_skips.exs` line 21 (and the test file): `collectgarbage("count")` asserts `{[0.0, 0], _}`. Matching on a literal `0.0` emits the Elixir "pattern matching on 0.0 is equivalent to matching only on +0.0" warning. Use `+0.0` to silence it and be explicit. CI runs plain `mix test` (no warnings-as-errors on test files) so it won't fail, but worth fixing.
2. The "no-ops returning 0" test groups `setpause`/`setstepmul` with the others. In real PUC-Lua those return the *previous* pause/stepmul value (a nonzero number), not 0 — and `count` returns real nonzero memory. These are genuine divergences but are filed under "stub behaviour" rather than the explicit "Divergence from PUC-Lua" group. Consider noting them so the future-real-collector checklist is complete.
3. Commit body prose says "Lines 1-165" while the skip starts at 167 (so 1–166 run; 166 is blank, 165 is `print("steps (2)")`). Harmless off-by-one in the prose only; the range is correct.

Nothing here blocks merge.

davydog187:
## Independent review — confirms the existing review; nothing blocking

Re-ran the full validation against the **committed** PR branch (FETCH_HEAD):

- `mix test` → 2044 passed, 21 skipped (matches PR claim).
- `mix test test/lua53_suite_test.exs --only lua53` → 15 passed, 14 skipped; gc.lua flips into the passing set. The run prints the prefix end-to-end (`tables` … `steps (2)` then `OK`), confirming line 624 `print('OK')` executes and the 167..622 region is correctly suppressed.
- `mix compile --warnings-as-errors` → clean.
- `collectgarbage_test.exs` → 6 passed.

**Boundary is correct.** `apply_skip_ranges` replaces each line in 167..622 with a comment (preserving line count), so `dosteps` (def at 167) and its call sites (186–191) are commented out together; nothing past 622 references a definition inside the range. The prefix (1–166) only touches stub-compatible modes (`collect`/`isrunning`/`setpause`/`setstepmul`/`step`), so it passes against the no-op stub. `collectgarbage"stop"` (182) is inside the skip, so it can't perturb the prefix.

**Conventions clean:** `test(suite):` scope on the implementation commit, `chore(A22a):` only on plan-only commits (allowed exception), `Closes #260` in body, no Co-Authored-By, no plan-id in source/tests/skip-reason. Skip reason precise with `issue: 260`.

**Notes (non-blocking), not duplicating the earlier review:**
1. The committed test already uses `+0.0` for the `count` assertion, so the earlier nit about the `0.0` pattern-match warning is already resolved on the branch (the `gh pr diff` rendering showing `0.0` is misleading).
2. The regression test pins *existing* stub behaviour, so it passes on `main` too — it does not fail pre-fix. That's expected and correct for a skip-narrowing triage PR (it's a documentation/checklist for the future collector, not a guard on a code fix). Worth being explicit that this isn't a bug-fix regression test.

LGTM.

PR #290, opened 2026-05-31
feat(vm): populate debug.getinfo name/namewhat from the call site

### LINKED ISSUE #279: debug.getinfo(level, 'n') returns nil for name and namewhat

davydog187:
## Goal

Make `debug.getinfo(level, "n")` return a non-nil `name` and a matching `namewhat`, so the issue repro passes:

```lua
function F(a) return a end
F(1)
assert(debug.getinfo(1, "n").name == 'F')
```

## Success criteria

- [x] `debug.getinfo(1, "n").name == "F"` for `function F() ... end; F()`
- [x] `namewhat` reports `global` / `local` / `upvalue` / `field` / `method` matching the call site, and `""` when no name is known
- [x] Regression unit test pins the issue repro
- [x] `constructs.lua` skip narrows past the `debug.getinfo` block (was `225..313`, now `232..313`)
- [x] `mix test` and the lua53 suite stay green

## Changes

The caller's call instruction already carries a compile-time name hint (`{:global, "F"}`, `{:local, "x"}`, ...) — the same hint that powers `attempt to call a nil value (global 'foo')` errors. Two gaps kept it from reaching `debug.getinfo`:

- The interpreter's `:call` handler pushed a `call_stack` frame for `:lua_closure` callees but routed `:compiled_closure` callees straight into `Dispatcher.execute` **without** recording a frame. A top-level `function F() ... end; F()` therefore ran with an empty `call_stack`, so `getinfo` had no name to report.
- `call_stack` frames stored only the textual name, not the hint tag needed to classify `namewhat`.

This PR:

- Records a `call_stack` frame for compiled-closure calls in the interpreter (mirroring the existing `:lua_closure` push), so the running function's name is always available.
- Threads the hint tag through to a new `namewhat` field on each frame, derived by `hint_namewhat/1` in `Lua.VM.Executor`.
- Reads the running function's frame (the head of `call_stack` when a native callback executes) in `Lua.VM.Stdlib.Debug` to surface `name`/`namewhat`.

No prototype/codegen change was needed: the call-site hint already lives on the instruction stream, which is the faithful PUC-Lua `getfuncname` model.

## Verification

```
mix format
mix compile --warnings-as-errors   # clean
mix test                           # 2040 passed, 22 skipped (was 2037)
mix test test/lua53_suite_test.exs --only lua53   # 14 passed, 15 skipped
```

## Out of scope

- The full PUC-Lua `getfuncname` runtime instruction walk for every form. We reuse the existing compile-time hint, which covers global / local / upvalue / field / method call sites.
- Naming of functions reached through a call site that carries no hint (call through a temporary, `(t[expr])()`, immediately-invoked anonymous closures) and functions reached by a tail call: these keep `name == nil` / `namewhat == ""`, matching PUC-Lua's unknown fallback.
- `name`/`namewhat` for the entry chunk and for native (C) functions inspected by value (`debug.getinfo(fn, "n")`).

Closes #279

### DIFF
--- a/.agents/plans/A44-debug-getinfo-name.md
+++ b/.agents/plans/A44-debug-getinfo-name.md
@@ -1,0 +1,89 @@
+---
+id: A44
+title: "debug.getinfo(level, 'n') populates name/namewhat from the caller's call site"
+issue: 279
+pr: null
+branch: feat/debug-getinfo-name
+base: main
+status: in-progress
+direction: A
+unlocks:
+  - constructs.lua
+---
+
+## Goal
+
+Make `debug.getinfo(level, "n")` return a non-nil `name` (and a matching
+`namewhat`) for the common call forms, so the issue repro passes:
+
+```lua
+function F(a) return a end
+F(1)
+assert(debug.getinfo(1, "n").name == 'F')
+```
+
+## Out of scope
+
+- The full PUC-Lua `getfuncname` caller-instruction walk for every form.
+  We reuse the compile-time name hint already threaded onto `:call`/`:self`
+  instructions (the same hint that powers "attempt to call a nil value
+  (global 'foo')" errors). That hint covers global, local, upvalue, field,
+  and method call sites.
+- Naming of functions reached through forms that carry no hint at the call
+  site (e.g. a call through a temporary, `(t[expr])()`, immediately-invoked
+  anonymous closures). Those keep `name == nil` / `namewhat == ""`, matching
+  the PUC-Lua "unknown" fallback closely enough for the suite.
+- `name`/`namewhat` for the entry chunk and for native (C) functions inspected
+  by value (`debug.getinfo(somefunc, "n")`): still nil. PUC-Lua also leaves
+  the value-based query without a name unless it can find a global binding.
+
+## Success criteria
+
+- `debug.getinfo(1, "n").name == "F"` for `function F() ... end; F()`.
+- `namewhat` is `"global"` / `"local"` / `"upvalue"` / `"field"` / `"method"`
+  matching the call site, and `""` when no name is known.
+- A regression unit test pins the issue repro.
+- `constructs.lua` skip range narrows past line 226.
+- `mix test` and the lua53 suite stay green.
+
+## Implementation notes
+
+The name is recovered from the *caller's* call instruction, PUC-Lua style,
+but resolved at compile time: every `:call`/`:self` instruction already
+carries a `name_hint` tagged tuple (`{:global, "F"}`, `{:local, "x"}`, ...),
+and the executor already records `name: hint_name(name_hint)` on each
+`call_stack` frame.
+
+Two small changes:
+
+1. Executor: also stash the hint *tag* on the frame as `:namewhat`
+   (`"global"`, `"local"`, `"upvalue"`, `"field"`, `"method"`), alongside the
+   existing `:name`. Done at both call-frame construction sites
+   (`do_execute` `:call` handler and `dispatcher_call_info/3`).
+2. `debug.getinfo`: for an integer `level`, read the callee frame for that
+   level (`call_stack` is headed by the running function when a native
+   callback is executing) and surface its `name`/`namewhat`.
+
+No prototype/codegen change is needed: the hint already lives on the
+instruction stream, which is the more faithful PUC-Lua model than capturing
+the declared name on the prototype.
+
+## Verification
+
+- `mix format`
+- `mix compile --warnings-as-errors`
+- `mix test`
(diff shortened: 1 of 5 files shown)


davydog187:
## Review: feat(vm): populate debug.getinfo name/namewhat from the call site

Reviewed for correctness, tests, conventions, and simplification. **Verdict: LGTM, ship it.** No blockers.

### What I verified
- **Regression tests genuinely pin the bug.** Ran the three new `debug_test.exs` cases against the pre-fix source (main's `debug.ex` + `executor.ex`): 3 of them fail (`name == nil` / all-nil namewhat), and they pass on the branch. The "no hint" case correctly passes both ways since it asserts `nil` / `""`.
- **Level-to-frame indexing is correct.** Name uses `call_stack[level-1]` while position uses `call_stack[level-2]` — these legitimately differ because a frame stores the *callee's* name (from the call instruction's `name_hint`) but the *caller's* line/source. Walked chunk→A→B→getinfo by hand; level 1/2/3 all resolve to the right name, namewhat, and currentline.
- **The compiled-closure frame push is sound.** `Dispatcher.do_execute_top` does not push a frame or bump `call_depth` on entry — it relies on callers to do so, exactly as the dispatcher's own internal call sites (push → execute → pop) do. The interpreter's `:call` handler previously skipped this for `:compiled_closure`, so the fix also (correctly) makes the interpreter→dispatcher hop respect `check_call_depth!`. Pop ordering mirrors the existing sites.
- **Full validation green:** `mix format --check-formatted`, `mix compile --force --warnings-as-errors` (clean), `mix test` (2040 passed, 22 skipped), `mix test --only lua53` (14 passed, 15 skipped).
- **Skip narrowing is minimal and real.** `constructs.lua` 225..313 → 232..313 un-skips the `debug.getinfo(1,"n").name == 'F'` block (lines 225-231), which now runs as part of the suite and passes. Reason string dropped the stale `debug.getinfo` mention; issue field stays `nil`.

### Conventions: all clean
- Title/feat scope is `vm` (allowed subsystem), not a plan id. `chore(A44)` commits touch **only** the plan file, which is the documented exception.
- No plan-id references in any source/test the PR adds.
- No Co-Authored-By trailers.
- Body has `Closes #279`.

### Minor (non-blocking)
- `hint_namewhat/1` has 2-tuple clauses `{:field, _}` and `{:method, _}`, but the compiler's `name_hint/2` only ever emits the 3-tuple forms `{:field, name, recv}` / `{:method, name, recv}`. The 2-tuple clauses are dead, though they harmlessly mirror the defensive style in `format_target_hint/1`.
- The value-based path (`debug.getinfo(somefunc, "n")`) still omits `namewhat` entirely, so it reads as `nil` in Lua rather than PUC-Lua's `""`. Pre-existing and explicitly out of scope here.

PR #291, opened 2026-05-31
fix(pattern): validate gsub replacement string and value

### LINKED ISSUE #261: Triage cluster — metamethod and control-flow assertion failures
Tracked by plan: [`.agents/plans/A23-triage-metamethod-control-flow.md`](https://github.com/tv-labs/lua/blob/main/.agents/plans/A23-triage-metamethod-control-flow.md)

When this work is picked up:

3. Merge closes this issue.

davydog187:
## Summary

`string.gsub` silently accepted replacement strings and replacement values that Lua 5.3 rejects. It now raises the Lua 5.3 errors:

- `invalid capture index %N in replacement string` for an out-of-range `%N` in the replacement string (e.g. `%2` against a pattern with no captures).
- `invalid use of '%' in replacement string` for a `%` followed by a non-digit, non-`%` character (only `%%` remains a literal percent).
- `invalid replacement value (a TYPE)` when a table or function replacement yields a non-string/number value.

The no-capture `%1` == `%0` quirk and the `%%` literal escape are preserved.

This advances the official `pm.lua` suite past its `checkerror` block (lines 233-238).

## Scope

In scope: gsub replacement-string and replacement-value validation.

Out of scope (narrowed to precise skip ranges in `test/lua53_skips.exs`, tracked under #261 for follow-up):
- Pattern backreferences `%0`/`%1` to an unclosed/absent capture (pm.lua 236-237).
- A pattern recursion/complexity limit for huge `.?` patterns (pm.lua 240-263).
- gsub-with-captures replacing only the capture, gmatch position captures, `%f` frontier sets, malformed-pattern error text, and embedded-`\0` patterns (pm.lua 277-371).

`pm.lua` moves from a whole-file `:all` skip to three narrowed ranges.

## Verification

- `mix compile --warnings-as-errors` — clean
- `mix test` — 2043 passed, 21 skipped
- `mix test test/lua53_suite_test.exs --only lua53` — 15 passed, 14 skipped

Plan: `.agents/plans/A23a-gsub-replacement-errors.md`

Closes #261

### DIFF
--- a/.agents/plans/A23a-gsub-replacement-errors.md
+++ b/.agents/plans/A23a-gsub-replacement-errors.md
@@ -2,10 +2,10 @@
 id: A23a
 title: "string.gsub validates replacement string and replacement value"
 issue: 261
-pr: null
+pr: 291
 branch: fix/metamethod-control-flow
 base: main
-status: in-progress
+status: review
 direction: A
 unlocks:
   - pm.lua
--- a/.agents/plans/A23a-gsub-replacement-errors.md
+++ b/.agents/plans/A23a-gsub-replacement-errors.md
@@ -1,0 +1,92 @@
+---
+id: A23a
+title: "string.gsub validates replacement string and replacement value"
+issue: 261
+pr: null
+branch: fix/metamethod-control-flow
+base: main
+status: in-progress
+direction: A
+unlocks:
+  - pm.lua
+---
+
+## Goal
+
+Make `string.gsub` raise the Lua 5.3 errors for an invalid replacement
+string or replacement value, so the official `pm.lua` suite advances
+past its `checkerror` block (lines 233-238).
+
+## Out of scope
+
+- Pattern-engine backreferences `%0` / `%1` to an unclosed or absent
+  capture (pm.lua 236-237) — needs capture-level validation in the
+  matcher.
+- A pattern recursion / complexity limit so huge `.?` patterns raise
+  "pattern too complex" (pm.lua 240-263).
+- `gsub` with captures replacing only the capture rather than the whole
+  match, plus gmatch position captures, `%f` frontier sets,
+  malformed-pattern error messages, and embedded-`\0` patterns
+  (pm.lua 277-371).
+- The other A23 cluster files (events.lua, errors.lua, closure.lua,
+  goto.lua).
+
+## Success criteria
+
+- [ ] `string.gsub` raises `invalid capture index %N in replacement string`
+      for an out-of-range `%N` in the replacement string.
+- [ ] `string.gsub` raises `invalid use of '%' in replacement string`
+      for a `%` followed by a non-digit, non-`%` char.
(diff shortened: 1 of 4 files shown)


davydog187:
## Review: fix(pattern) gsub replacement validation

Verified the behaviour against a reference Lua interpreter and read the full diff plus the gsub dispatch in `lib/lua/vm/stdlib/string.ex`. This is a clean, well-scoped fix.

### Correctness — looks right
- `%2` against a no-capture pattern, lone `%x` escape, `%%` literal (count 2), and non-string/number table/function results all match reference Lua semantics. Spot-checked each against PUC-Lua.
- The `replace_captures/3` `cond` is sound: `idx == 0` -> whole; `idx <= length(captures)` -> capture (covers position/integer captures via `capture_to_binary`); `idx == 1` only reachable with zero captures (the documented no-capture quirk -> whole); else raise. Dropping the old `Enum.at(...) || ""` fallback is safe since captures are always binaries or integers, never nil.
- Both table and function replacements funnel through the arity-2 `is_function` branch in `string.ex`, so the `invalid replacement value` check covers both — the tests confirm this.
- The `"invalid capture index %N in replacement string"` wording keeps the 5.3 suffix (5.4+ dropped it); correct for a 5.3 VM.

### Tests — pin the bug
- The four error-path tests assert `ok == false`; on `main` these inputs silently succeed (`%2` -> `""`, `%x` -> `x`), so they fail pre-fix. Good regression coverage. The `%%` test pins the preserved literal-escape behaviour.

### Skips — minimal and precise
- `pm.lua` moves from a whole-file `:all` skip to three narrowed ranges (236-237, 240-263, 277-371), each with a precise reason and `issue: 261`, no plan-id. Confirmed line 238 (`%x`) sits between skip ranges and now executes/passes in the suite, and 233-235 run too. `--only lua53` stays green (15 passed, 14 skipped).

### Conventions — clean
- `fix(pattern): ...` subsystem scope; the two `chore(A23a)` commits each touch only the plan file (the allowed exception). No `Co-Authored-By` trailers. Body has `Closes #261`. No plan-id in source or tests.

### Minor (non-blocking)
- There is a dedicated terminal clause `replace_captures("%", ...)` for a trailing lone `%`, but no regression test exercises it specifically (the `%x` test hits the other lone-`%` clause). Worth a one-line case if you want full coverage.

Validation re-run locally: `mix format --check-formatted`, `mix compile --warnings-as-errors`, `mix test` (2043 passed, 21 skipped), and `--only lua53` (15 passed, 14 skipped) all green.

LGTM.

davydog187:
Added a focused regression for the trailing lone `%` terminal clause: `string.gsub("alo", ".", "x%")` now asserts it raises `invalid use of '%' in replacement string`. This exercises the `replace_captures("%", ...)` path distinctly from the existing `%x` lone-% test. Full suite green (2076 passed).

PR #293, opened 2026-05-31
fix(stdlib): table.unpack rejects oversized ranges

### LINKED ISSUE #262: Triage cluster — stdlib and data-structure assertion failures
Tracked by plan: [`.agents/plans/A24-triage-stdlib-data-structure.md`](https://github.com/tv-labs/lua/blob/main/.agents/plans/A24-triage-stdlib-data-structure.md)

When this work is picked up:

3. Merge closes this issue.

davydog187:
## Goal

`table.unpack(list, i, j)` built the entire `i..j` slice before
returning, so a large `j` (e.g. `math.maxinteger`) sent the VM into a
multi-billion-element loop. Lua 5.3's `ltablib.c` checks the range up
front and raises `too many results to unpack` when the element count
`(j - i) + 1` reaches `INT_MAX` (its count-overflow and `lua_checkstack`
arms both reduce to this). This adds that guard before any slot is read.

This is the single most tractable concrete failure triaged out of the
A24 stdlib/data-structure cluster (#262): the first assertion `sort.lua`
hits, at line 16, is driven by `checkerror("too many results", unpack,
{}, 0, maxi)`.

## Changes

- `lib/lua/vm/stdlib/table.ex`: `table_unpack/2` raises
  `Lua.VM.RuntimeError` with `too many results to unpack` when
  `i <= j and j - i >= 2_147_483_646` (count `>= INT_MAX`), before
  materialising the slice. This models both PUC rejection arms, including
  the `lua_checkstack` arm that fires at count exactly `INT_MAX`. The
  subtraction uses Elixir bignums, so it cannot overflow.
- `test/lua/vm/stdlib/table_test.exs`: regression coverage for the
  rejected range, the `(1 << 31) - 1` boundary, the count-exactly-INT_MAX
  case, a normal range, and a reversed (empty) huge range.
- `test/lua53_skips.exs`: `sort.lua` stays a whole-file skip — its first
  assertion no longer fails, but the file still times out on the
  O(n^2) sort sections and the unimplemented `os.clock` timing harness.
  The skip reason is updated and now carries `issue: 262`.

## Out of scope

- `constructs.lua` os/debug/short-circuit cases (other PRs in the batch).
- `literals.lua` parse-error wording / `debug.getinfo` line skips.
- `table.sort` "too big" / "invalid order function" detection and the
  `table.move` "materialise-before-write" interruption semantics
  (sort.lua line 174) — separate, larger follow-ups.

## Verification

- `mix format`, `mix compile --warnings-as-errors` clean.
- `mix test`: 2040 passed, 22 skipped, 0 failures.
- `mix test test/lua53_suite_test.exs --only lua53`: 14 passed, 15
  skipped, 0 failures (no lua53 delta — `sort.lua` was and remains a
  whole-file skip; the unpack correctness fix is proven by the new unit
  tests).

Plan: A24a

Closes #262

### DIFF
--- a/.agents/plans/A24a-table-unpack-too-many-results.md
+++ b/.agents/plans/A24a-table-unpack-too-many-results.md
@@ -1,0 +1,76 @@
+---
+id: A24a
+title: "table.unpack raises 'too many results' for out-of-range slices"
+issue: 262
+pr: null
+branch: fix/stdlib-data-structure
+base: main
+status: in-progress
+direction: A
+---
+
+## Goal
+
+Make `table.unpack(list, i, j)` reject ranges that would produce more
+than `INT_MAX` results (raising `too many results to unpack`) before it
+tries to materialise the slice, matching Lua 5.3 `ltablib.c`.
+
+This is the single most tractable concrete failure in the A24 cluster
+(parent plan `.agents/plans/A24-triage-stdlib-data-structure.md`,
+issue #262): the first assertion `sort.lua` hits is at line 16, driven
+by `checkerror("too many results", unpack, {}, 0, maxi)` on line 48.
+
+## Out of scope
+
+- `constructs.lua` os/debug/short-circuit cases (other PRs in the batch).
+- `literals.lua` parse-error wording and `debug.getinfo` line skips.
+- `table.sort` "too big" / "invalid order function" checks (sort.lua
+  lines 199, 204) — those remain behind the narrowed skip range.
+- The `os.clock`-driven timing section of `sort.lua`.
+- Stable/quadratic sort behaviour and performance.
+
+## Success criteria
+
+- [x] `table.unpack({}, 0, math.maxinteger)` raises an error whose
+      message contains `too many results` instead of hanging.
+- [x] Small/valid `table.unpack` ranges are unaffected.
+- [x] Regression test under `test/lua/vm/` covering the boundary.
+- [x] `mix test` passes.
+- [x] `mix test test/lua53_suite_test.exs --only lua53` passes.
+
+## Implementation notes
+
+- `lib/lua/vm/stdlib/table.ex` `table_unpack/2`: after resolving `i`/`j`
+  and the integer checks, if `i <= j and j - i >= INT_MAX` raise
+  `Lua.VM.RuntimeError` with `too many results to unpack`. The
+  subtraction is on Elixir bignums so it cannot overflow.
+- `INT_MAX` is `2_147_483_647`; PUC rejects `(j - i) >= INT_MAX`.
+
+## Verification
+
+```bash
+mix format
+mix compile --warnings-as-errors
+mix test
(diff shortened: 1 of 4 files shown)


davydog187:
## Review — approve with one nit

Verified locally on `fix/stdlib-data-structure`: `mix test` → 2040 passed, 22 skipped, 0 failures; `mix test test/lua53_suite_test.exs --only lua53` → 14 passed, 15 skipped, 0 failures. Matches the PR claims.

### Correctness
The guard `i <= j and j - i >= 2_147_483_647` faithfully mirrors PUC `ltablib.c`:
```c
if (i > e) return 0;            /* empty range, checked first */
n = (lua_Unsigned)e - i;       /* count - 1 */
if (n >= (unsigned int)INT_MAX || !lua_checkstack(...)) luaL_error(...);
```
The `i <= j` precondition correctly preserves the empty-range short-circuit, and bignum subtraction can't overflow. The error value `"too many results to unpack"` and the `RuntimeError.exception(value:)` surfacing pattern match the rest of stdlib. Boundary case `(1 << 31) - 1` is rejected on both sides — correct.

### Tests
Cover the rejected range, the `INT_MAX` boundary, a normal range, and a reversed (empty) huge range. The pre-fix path would hang on `math.maxinteger`, so a guard (rather than a slower materialize loop) is the right shape. Good coverage.

### Conventions
Clean. `fix(stdlib):` subsystem scope; the two `chore(A24a):` commits touch only the plan file (the allowed exception); no plan-id in source/tests/comments in this diff; no Co-Authored-By trailers in commits; body has `Closes #262`. The `sort.lua` skip stays whole-file but the reason is now precise and carries `issue: 262` — reasonable since the file times out on O(n^2) sort + unimplemented `os.clock`, so a line-range skip wouldn't help.

### Minor
- **Edge (minor):** PUC also rejects a count of exactly `INT_MAX - 1` via the `lua_checkstack(++n == INT_MAX)` arm; this PR's guard only fires at `>= INT_MAX`, so a `j - i == INT_MAX - 1` range (~2.1B elements) would still be materialized rather than rejected. Practically unreachable and outside this fix's target assertion — fine to leave.
- **Nit:** PR body carries the `🤖 Generated with Claude Code` footer. Commits are clean; the explicit Co-Authored-By prohibition is satisfied, but the footer is adjacent to the plain-authorship convention.

LGTM.

davydog187:
Both review items are resolved:

- **Minor edge — INT_MAX-1:** the guard is tightened to `j - i >= @unpack_max_results - 1` (table.ex:414), so a count of exactly `INT_MAX` (i.e. `j - i == INT_MAX - 1`) is now rejected, matching PUC's `lua_checkstack(++n == INT_MAX)` arm.
- **Nit — PR-body footer:** the `Generated with Claude Code` footer has been removed from the body.

Branch is mergeable/clean against current main.

PR #294, opened 2026-05-31
test(vm): pin short-circuit depth-4 semantics; correct constructs skip

### LINKED ISSUE #281: constructs.lua short-circuit harness fails at level=4 deep composition

davydog187:
## Summary

Bisected the `constructs.lua` short-circuit harness (lines 287-298) at
`level=4`, as flagged in #281. The suspected executor short-circuit bug
(register aliasing under conditional-jump bytecode, or a `not` precedence
wrinkle) **does not reproduce**.

A faithful in-process replica of the harness — `createcases` up to
`level=4`, the `if %s then IX = true end; return %s` program template, run
through `load` — passes **all 204105 cases** for both `_ENV.GLOB1 = 0` and
`_ENV.GLOB1 = 1`. The `test_and`/`test_or` CPS path in the executor
evaluates deep `((((a op b) op c) op d) op e)` compositions and their
`not(...)` wrappings correctly.

## Reduced findings

The real blockers gating `constructs.lua` lines 225-313 are upstream of the
harness and unrelated to short-circuiting:

1. **Line 226** `assert(debug.getinfo(1, "n").name == 'F')` — `debug.getinfo`
   always returns `name = nil`; call-stack name introspection is
   unimplemented.
2. **Line 237** `_ENV.GLOB1 = math.floor(os.time()) % 2` — `os.time` is nil
   (the `os` library is essentially unimplemented). `GLOB1` then feeds line
   248's concat and the harness leaf `(0==_ENV.GLOB1)`, so line 237 cannot
   simply be blanked.

Both are out of scope for this issue, which targeted the short-circuit
executor path. This is therefore the "otherwise" branch of #281: no
tractable executor fix, so the skip reason is corrected and the harness is
pinned green by a regression test.

## Changes

- `test/lua/vm/short_circuit_test.exs` — new regression test pinning the
  short-circuit semantics: explicit depth-4 `and`/`or` chains, `not`
  wrapping, the if-condition side-effect shape, and the level-3 generative
  harness (4105 cases). The full level-4 run (204105 cases) passes
  identically but is too slow for the default suite.
- `test/lua53_skips.exs` — `constructs.lua` skip reason rewritten to name
  the two upstream blockers and reference the issue; previously it wrongly
  implicated a "load()-driven short-circuit harness" edge case.

## Why draft

Per #281, the directive was to fix the executor bug if tractable, otherwise
narrow the skip with a precise reason and open a draft documenting the
reduced repro. The executor is correct, so there is no fix to land; this PR
documents that and pins it. Opened as draft for review of the
classification before the skip-reason correction merges.

## Validation

- `mix format`
- `mix compile --warnings-as-errors`
- `mix test` — 2051 passed, 22 skipped
- `mix test test/lua53_suite_test.exs --only lua53` — 14 passed, 15 skipped (no delta; constructs.lua stays skipped with an accurate reason)

Plan: A45

Closes #281

### DIFF
--- a/.agents/plans/A45-short-circuit-level4.md
+++ b/.agents/plans/A45-short-circuit-level4.md
@@ -2,10 +2,10 @@
 id: A45
 title: "Bisect constructs.lua short-circuit harness at level=4; pin green, narrow skip"
 issue: 281
-pr: null
+pr: 294
 branch: fix/short-circuit-level4
 base: main
-status: in-progress
+status: review
 direction: A
 unlocks:
   - constructs.lua
--- a/.agents/plans/A45-short-circuit-level4.md
+++ b/.agents/plans/A45-short-circuit-level4.md
@@ -1,0 +1,85 @@
+---
+id: A45
+title: "Bisect constructs.lua short-circuit harness at level=4; pin green, narrow skip"
+issue: 281
+pr: null
+branch: fix/short-circuit-level4
+base: main
+status: in-progress
+direction: A
+unlocks:
+  - constructs.lua
+---
+
+# Goal
+
+Bisect the dynamic short-circuit harness at `constructs.lua:287-298` at
+`level=4`, find the smallest failing `((((a op b) op c) op d) op e)`
+composition of `and`/`or`, reduce to a one-line repro, and classify the
+suspected executor short-circuit edge case (register aliasing under
+conditional-jump bytecode, or a `not` precedence wrinkle). Fix the
+executor bug if tractable and land green; otherwise narrow the
+`constructs.lua` skip range with a precise reason + issue and document
+the reduced repro.
+
+# Out of scope
+
+- Implementing the `os` library (`os.time`).
+- `debug.getinfo(..., "n").name` call-stack name introspection.
+- Any feature beyond the short-circuit executor path.
+
+# Findings
+
+The suspected executor short-circuit bug does **not** reproduce. A faithful
+replica of the harness — `createcases` up to `level=4`, the
+`if %s then IX = true end; return %s` program template, run through the
+suite-runner `load` — passes **all 204105 cases** for both `_ENV.GLOB1 = 0`
+and `_ENV.GLOB1 = 1`. The `and`/`or` `test_and`/`test_or` CPS path in the
+executor evaluates deep compositions correctly, including the
+`not(<expr>)` wrapping the harness applies.
+
+The real blockers gating lines 225-313 are upstream of the harness and
+unrelated to short-circuiting:
+
+1. **Line 226** `assert(debug.getinfo(1, "n").name == 'F')` — the VM's
+   `debug.getinfo` always returns `name = nil`; call-stack name
+   introspection is unimplemented.
+2. **Line 237** `_ENV.GLOB1 = math.floor(os.time()) % 2` — `os.time` is
+   nil (the `os` library is essentially unimplemented). `GLOB1` then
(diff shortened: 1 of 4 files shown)


davydog187:
## Review: sound investigation, ready to land green

I independently reproduced the central claim. Running the full **level-4** harness (204105 cases) in-process with a faithful `load` install passes completely — no executor short-circuit bug. I also confirmed both upstream blockers are real:

- `os` global is nil, so `os.time` at line 237 raises before the harness runs.
- `debug.getinfo(1,"n").name` returns nil, and the `assert(... == 'F')` in `F` (called at line 230) fires an assertion failure well before the harness at 287.

So the harness genuinely never executes in the suite, and the previous "load()-driven short-circuit harness" skip reason was indeed a misattribution. The classification holds: this is the "otherwise" branch of #281 — no tractable executor fix, corrected skip reason, harness pinned by regression test. Issue #281's described "fails at ~120000 iterations" symptom does not reproduce on current main (likely fixed by intervening closure/upvalue work, or it was always the blockers masking the harness).

**Conventions:** all clean. `test(vm):` subsystem scope; `chore(A45)` only on plan-only commits (the allowed exception); no plan-id in source/tests/skip-reason; no Co-Authored-By trailer; body has `Closes #281`. Skip reason is precise and references the issue.

**Tests:** the 14 cases pass; hand-computed expected values spot-checked correct, including the `error()` short-circuit-order cases and the if-condition side-effect shape. Generative level-3 harness (4105 cases) is self-contained on default `load`.

### Minor notes (non-blocking)
1. The regression test pins **level 3** (4105 cases) for suite speed; level 4 (the level #281 actually flagged) is only verified by the manual run in the PR narrative, not in CI. Consider a `@tag :slow` level-4 variant excluded by default so the exact regression surface is reproducible on demand rather than living only in prose.
2. The test module comment mentions guarding against "register aliasing" and a "`not` precedence wrinkle" — but the PR's own conclusion is that neither exists. Reads slightly as if pinning a known hazard; a one-line note that these were the *suspected* (and disproven) mechanisms would keep the comment honest.

Nothing here blocks marking ready once `mix test` and the `--only lua53` suite are green (PR reports both pass with no suite delta).

PR #304, opened 2026-05-31
feat(error): lead rendered errors with location and gate ANSI on TTY

### LINKED ISSUE #263: Error-message quality pass — audit rendered output against PUC-Lua
Tracked by plan: [`.agents/plans/A26-error-message-quality.md`](https://github.com/tv-labs/lua/blob/main/.agents/plans/A26-error-message-quality.md)

When this work is picked up:

3. Merge closes this issue.

davydog187:
## Error message quality pass — make the rendered output world-class

Plan: [`.agents/plans/A26-error-message-quality.md`](https://github.com/tv-labs/lua/blob/errors/quality-pass/.agents/plans/A26-error-message-quality.md)
Closes #263

### Goal

A18/A19 wired the location and structured-error *data* into every VM
exception. This pass audits the *rendered* output a user actually sees and
fixes the rough edges: location prominence, a redundant double label, raw
ANSI leaking off a TTY, and dead/filler suggestions.

### Success criteria

- [x] The rendered output for each error category is pinned by `test/lua/error_gallery_test.exs`, with an inline `expected` string per case (arithmetic, index nil, call nil, concat, compare, stdlib bad-arg, `assert(false[, msg])`, `error("...")`, `error({tbl})`, stack overflow). Verified: `mix test test/lua/error_gallery_test.exs`.
- [x] Each message leads with `at <source>:<line>:` before the body — reordered in `ErrorFormatter.format/3`; pinned by the golden tests in `error_to_map_test.exs` and every gallery case. Because `Lua.RuntimeException` prefixes `Lua runtime error: `, the first rendered line is `Lua runtime error: at <source>:<line>:`.
- [x] Double label gone — the formatter no longer emits a "Runtime Type Error" header; the public `Lua runtime error:` prefix is the single label. Verified in the gallery cases.
- [x] Suggestions are category-specific and reachable — the dead `:arithmetic_type_error` clause is realigned to the emitted `:arithmetic_on_non_number`; added suggestions for `:compare_incompatible_types`, `:length_not_integer`, `:bitwise_on_non_integer`, `:for_loop_non_number`; removed the filler assertion suggestion.
- [x] ANSI gated on `IO.ANSI.enabled?/0` — no escape codes leak when piped. Verified: piping `Lua.eval!` output to a file contains zero `\e[` sequences.
- [x] Error gallery pins before/after for each category — shipped as inline `expected` strings in `test/lua/error_gallery_test.exs` (now including `bitwise_on_non_integer` and `for_loop_non_number`). The standalone `guides/errors.md` draft was dropped during review; the known-gaps notes live in this PR description and the plan's Discoveries.
- [x] `mix test` passes — 2104 passed, 0 failures (was green before; the formatter golden tests were updated to the new format).

### Changes

\`\`\`
 .agents/plans/A26-error-message-quality.md | 308 +++++++++++++++++++++++------
 lib/lua/vm/error_formatter.ex              |  98 +++++----
 lib/lua/vm/runtime_error.ex                |  11 +-
 test/lua/error_gallery_test.exs            | 144 ++++++++++++++
 test/lua/error_messages_test.exs           |   6 +-
 test/lua/vm/error_to_map_test.exs          |  51 +++--
\`\`\`

### Discoveries

- A19's location data is present on this base; every reachable error carries a location except two data-layer-gap cases: stack overflow, and the numeric `for`-loop control coercion (`for_loop_non_number`). The latter's `raise` in `coerce_for_value/2` receives no position from `current_position/0`, so it renders with no `at <source>:<line>:` header — identical on `main`, not a regression — and the gallery honestly pins the location-less output.
- Out-of-scope data-layer gaps (logged, not fixed): `#5`/`#true` return `0` instead of raising; `t[nil]=1`/`t[0/0]=1` succeed silently; stack-overflow errors and numeric `for`-loop control coercion errors carry no originating line. These are documented under "Known gaps" in the plan's Discoveries section (the standalone `guides/errors.md` draft was dropped during review).
- `error()` on a non-string now renders PUC-Lua's `(error object is a TYPE value)` instead of leaking `{:tref, 12}`.
- `assert(false)` still double-texts (`assertion failed: assertion failed!`); removing the prefix ripples into the official suite + integration tests, so it is left as a follow-up (the plan scoped the assertion change to its suggestion, now removed).
- `error_to_map_test.exs` / `error_messages_test.exs` golden snapshots were updated as a direct consequence of the formatter change.

### Verification

\`\`\`
mix format
mix test                 # 2104 passed, 19 skipped, 0 failed
mix test --only lua53    # 17 passed, 0 failed
\`\`\`

Manual non-TTY leak check: piping a rendered error to a file contains no raw `\e[` escape sequences.

- Changing the exception struct shape or adding new error categories.
- Implementing `traceback()` / `debug.traceback`; source maps past `load()` chunks.
- The two A19-flagged data oddities and the data-layer gaps above.


guides/errors.md
```diff
@@ -0,0 +1,210 @@
...
+## Known gaps (tracked separately)
+
+A few categories do not yet render ideally because the *data* feeding the
+renderer is incomplete — these are data-layer issues, not rendering ones:
+
+- The length operator on a non-string/table (`#5`) and assigning a `nil`/NaN
+  table key (`t[nil] = 1`) do not currently raise, so they have no rendered
+  message to show.
+- The stack-overflow runtime error carries no originating line, so it renders
+  without an `at <source>:<line>:` header and its frames show line `0`.
```
davydog187: where are these tracked?

test/lua/error_gallery_test.exs
```diff
@@ -0,0 +1,96 @@
...
+    {"call_nil", "local f = nil\nf()", []},
+    {"concat_non_string", "local t = {}\nprint(t .. \"x\")", []},
+    {"compare_incompatible", "print(1 < \"x\")", []},
+    {"stdlib_bad_arg", "string.upper(nil)", []},
+    {"assert_with_message", "assert(false, \"boom\")", []},
+    {"assert_no_message", "assert(false)", []},
+    {"error_string", "error(\"something broke\")", []},
+    {"error_table", "error({code = 1})", []},
+    {"stack_overflow", "local function f(n) return 1 + f(n + 1) end\nf(1)", [max_call_depth: 30]}
+  ]
```
davydog187, line 133: Is this really the best way to do this, as opposed to using ~LUA and having them local? seems like a lot of complexity here for minimal gain

davydog187:
## Automated review — round 1

Solid pass. The formatter change is clean, ANSI is now fully routed through the `color/2` helper (no escape leaks off a TTY), the golden snapshots and inline gallery are updated to the new format, and the full suite is green locally (2104 passed). The `error()` non-string rendering now matches PUC-Lua (`(error object is a TYPE value)`), including the correct handling of `nil`/boolean. Stack-overflow truncation math checks out and the `max_call_depth: 30` makes that fixture deterministic.

A few findings:

### Major — `:length_not_integer` suggestion is misleading
`lib/lua/vm/error_formatter.ex` adds:
> The length operator (#) works on strings and tables. Use it on a string or table value.

But the only site that emits `:length_not_integer` (`lib/lua/vm/executor.ex:2382-2392`) raises **"object length is not an integer"** — i.e. a table length / `__len` metamethod that returned a non-integer (float/string). The user already applied `#` to a table; telling them to "use it on a string or table value" contradicts the message body. (`#5` on a number doesn't even raise — it's a documented Known gap.) The suggestion should speak to a length result that isn't an integer, e.g. "the table's length (or its `__len` metamethod) produced a non-integer value."

### Minor — plan/PR claim fixture files that no longer exist
The PR body and `.agents/plans/A26-error-message-quality.md` (success criteria + "What changed", lines ~316-317 and ~136/177) repeatedly assert `test/fixtures/error_gallery/*.txt` (11 files) exist and were verified. The final commit `eb8e87d` deliberately dropped them in favor of inline `expected` strings in `error_gallery_test.exs` (a good call). The committed plan is now factually wrong about its own deliverable — please update those references.

### Minor — location is not actually "on its own line"
The success criterion says each message leads with `at <source>:<line>:` *on its own line*. In practice the formatter output starts with `at ...:` (no leading newline), and `Lua.RuntimeException` prepends `Lua runtime error: `, so it renders as `Lua runtime error: at gallery.lua:2:` on a single line. The concrete before/after examples in `guides/errors.md` show this honestly, but the abstract "layout" block at the top implies a standalone location line. Either reword the criterion/guide or split the line — worth a deliberate decision.

### Nit — commit scope uses plan id
`chore(A26): start plan` / `chore(A26): mark plan as review` use the plan id as scope; the convention's allowed scope is `chore(plan):`. (There is mixed precedent on `main`, and history can't be rewritten now — noting for next time. The PR title itself is correctly scoped.)

Nothing here blocks merge except the `:length_not_integer` suggestion, which ships incorrect user-facing guidance.

davydog187:
## Automated review — round 2

Re-reviewed the pushed tip (`3ead830`) in a detached worktree off `origin/errors/quality-pass`. Ran `mix format --check-formatted`, `mix test` (2105 passed, 19 skipped, 0 failed), and `mix test --only lua53` (17 passed). All green.

The round-1 items look addressed: the dead `:arithmetic_type_error` clause is realigned to `:arithmetic_on_non_number`; ANSI is gated through the `color/2` helper on `IO.ANSI.enabled?/0`; location leads the render; the double "Runtime Type Error" header is gone; `length_not_integer` suggestion text now matches between code, gallery, and `guides/errors.md`. The gallery tests pin the new plain-text render and would fail against the pre-fix output (header + ANSI), so they pin real behavior.

### Findings

**1. (major) Plan-id used as commit scope.** Two commits on the branch use `chore(A26):` as the conventional-commit scope:

```
chore(A26): mark plan as review
chore(A26): start plan
```

The repo convention is explicit that the commit scope is the affected **subsystem, never a plan id** — `plan` is the allowed scope for plan-file housekeeping. These should be `chore(plan):`. The plan id belongs in the commit body (`Plan: A26`), not the scope. (Not force-fixing history is also a rule, so this is a note for the author / future commits rather than a request to rewrite the branch.)

**2. (nit, pre-existing — not introduced here) Source-context `^` pointer renders on the same line as the code.** In `render_source_context/1` the highlight branch emits `"\n" <> line_str` followed by `pointer` with no separating newline, so with ANSI disabled the golden snapshot is `   2 │ x()       ^` rather than the pointer sitting on its own line under the offending token. This predates the PR (the body is unchanged from `main`) and is out of scope, but since this PR's whole point is render quality it's worth a follow-up.

No correctness issues found in the in-scope changes. `error()` on non-string/number now renders `(error object is a TYPE value)` for nil/boolean/table/function/userdata and verbatim for strings/numbers, matching PUC-Lua; `String.trim_leading/1` correctly drops the leading indent only on the location-less (stack-overflow) path so the `Lua runtime error: ` prefix composes cleanly.

*Automated round-2 review.*

davydog187:
## Automated review — round 3

Reviewed the current remote tip (`1fe8d05`) in a detached worktree. Ran `mix format --check-formatted` (clean), `mix compile --warnings-as-errors` (clean), full `mix test` (2105 passed, 19 skipped, 0 failed), and `mix test --only lua53` (17 passed). The round-3 changes (caret on its own line, `length_not_integer` suggestion wording, inline gallery) all land cleanly and the golden snapshots match the actual render.

Verified independently:
- All five realigned/added suggestion kinds (`arithmetic_on_non_number`, `compare_incompatible_types`, `length_not_integer`, `bitwise_on_non_integer`, `for_loop_non_number`) are genuinely emitted in `executor.ex` — the previously-dead `:arithmetic_type_error` clause is correctly retired and the new clauses are reachable (triggered `1.5 & 2` and `for i=1,"x"` end-to-end and confirmed the suggestions surface).
- ANSI gating works both ways: with `ansi_enabled: true` the location/caret/stack all colorize; with it disabled (test env) `refute String.contains?(output, @ansi_escape)` holds.
- `error({tbl})` / non-string error objects now render PUC-Lua's `(error object is a table value)` via `Value.type_name/1` instead of leaking `{:tref, _}`. This also matches PUC's `lua.c` msghandler behavior for `nil`/boolean/function objects.

### Findings (all minor / non-blocking)

**1. `RuntimeError.stringify/1` formats error-object numbers with Elixir `to_string/1` instead of the canonical Lua formatter.** `lib/lua/vm/runtime_error.ex` does `stringify(v) when is_integer(v) or is_float(v), do: to_string(v)`. The rest of the codebase renders Lua numbers through `Lua.VM.Value.to_string/1` (Lua's `%.14g`-style formatting, e.g. `1.0e20` → `1e+20`). For `error(1.0e20)` this diverges from Lua. Integers and small floats are unaffected, so this is an edge case, but reusing `Lua.VM.Value.to_string/1` would keep number rendering consistent and correct for free.

**2. `guides/errors.md` "Known gaps" understates the missing-location cases.** The guide lists stack-overflow and the length/nil-key non-raising cases, but `for i=1,"x"` (and other `for_loop_non_number` paths) also render *without* an `at <source>:<line>:` header — the raise site in `executor.ex` doesn't carry line/source. This is a pre-existing data-layer gap (not introduced here), but the gallery's "every error leads with location" framing has more exceptions than the guide admits. Worth a one-line addition so the "world-class" claim stays honest.

**3. Commit-scope convention.** Two commits use `chore(A26): …` as the scope. The repo's non-negotiable convention is scope = subsystem, never a plan id (`chore(plan): …` is the allowed form). History is already pushed so this is noted for future hygiene rather than a fix request — do not force-push a ready PR to correct it.

No correctness blockers. The rendering reorder, double-label removal, ANSI gating, and suggestion realignment are all correct and well-pinned.

guides/errors.md
davydog187, line 1: we don't need a guide for error messages. We're also comparing it before and after a PR on this, this is not needed.

davydog187:
## Automated review — round 1

Pulled the current branch tip (`54ac0a2`), ran `mix format --check-formatted`, `mix compile --warnings-as-errors`, `mix test` (2106 passed, 19 skipped, 0 failed), and `mix test --only lua53` (17 passed). All green. The code change is solid: location-first ordering, the `color/2` gate, the realigned `:arithmetic_on_non_number` clause, and the four added (and executor-reachable) suggestion kinds all check out. The gallery tests would genuinely fail against the old format (old golden had `\e[31m...Runtime Type Error\e[0m`), so they pin the new behavior. The caret-on-its-own-line fix and `(error object is a TYPE value)` rendering match PUC-Lua.

A few non-code items:

### Major — commit scope uses the plan id, not a subsystem
Two commits violate the non-negotiable scope rule (subsystem, never a plan id):
- `4220888 chore(A26): mark plan as review`
- `80786e8 chore(A26): start plan`

Allowed scope here is `plan` (e.g. `chore(plan): mark A26 as review`). The `Plan: A26` trailer in the bodies is correct and should stay; only the subject scope is wrong.

### Minor — PR body and plan file still claim `guides/errors.md` ships
The final commit (`54ac0a2`) removed `guides/errors.md`, but the PR body success criteria ("Error gallery lives in `guides/errors.md`") and the plan file (`.agents/plans/A26-error-message-quality.md` lines 92, 174, 269, 318: "Files touched: `guides/errors.md` (new)", "documented under Known gaps") still describe it as delivered. The "Known gaps" content now lives only in the plan/PR discoveries. Update the criteria and file list so the record matches what shipped.

### Nit — two new suggestion kinds aren't pinned end-to-end
`:bitwise_on_non_integer` and `:for_loop_non_number` suggestions were added and are reachable (executor emits them at executor.ex:2757/2768/2784/2802/2818), but the gallery only pins `compare`/`length` (and `length` only via `to_map`). Optional: add gallery cases (`#true` won't work since `#` on non-table is a data-layer gap, but `1 & "x"` / `for i=1,"x" do end` should hit them) so a regression in those clauses is caught.

No correctness blockers. Resolve the commit-scope issue before merge.

davydog187:
## Automated round-2 review

Re-reviewed the current remote tip (`5b5b31e`). Validation all green from a detached checkout:

- `mix format --check-formatted` clean
- `mix compile --warnings-as-errors` clean
- `mix test` — 2108 passed, 19 skipped, 0 failed
- `mix test --only lua53` — 17 passed, 0 failed

The substance is solid: location-first ordering, the single-label resolution, the `color/2` ANSI gate (verified the `to_map/3` wire path never touches `IO.ANSI`), the realigned/added suggestions (all reachable — `:length_not_integer` fires from a non-integer `__len`, not dead), and the `RuntimeError.stringify/1` change rendering `(error object is a TYPE value)`. The latter matches the existing suite expectation at `test/lua53_tests/main.lua:253` and PUC-Lua's standalone message handler. Lua 5.3 §-semantics for each pinned message check out (`'for' limit must be a number`, `attempt to compare number with string`, bitwise-on-string, etc.). Gallery cases are deterministic and would fail before the fix (they pin exact output incl. the dropped header and stripped ANSI).

Only minor items, none blocking the behavior:

1. **Convention (history): two commits scope on the plan id** — `chore(A26): start plan` and `chore(A26): mark plan as review`. Repo rule says commit scope is the subsystem, never a plan id; plan-status chores should be `chore(plan): ...`. Already pushed and no-force-push/no-rewrite here, so flagging for future hygiene rather than asking for a rewrite.

2. **PR-description claim is slightly off** — "every reachable error carries a location except the stack-overflow case." The `for_loop_non_number` gallery case also renders with no `at <source>:<line>:` header (its raise in `coerce_for_value/2` gets no position from `current_position/0`). This is pre-existing data-layer behavior, not a regression, and the gallery honestly pins the location-less output — but the description should add for-loop to the "no location" list (or note it as another data-layer gap follow-up).

3. **nit — stale historical comment.** `test/lua/vm/for_loop_register_test.exs:15` still describes the old `Runtime Type Error — ...` render in a comment. Now that the header is gone this is mildly misleading; a one-line touch-up would keep it honest (not in this PR's diff, optional).

4. **nit — message-indent inconsistency.** Located cases render the body with a 2-space indent (`  attempt to ...`); location-less cases (`for_loop_non_number`, `stack_overflow`) get the indent stripped by `String.trim_leading/1` so the body sits flush-left. Deliberately pinned, but the two layouts differ. Acceptable as-is.

Nothing here blocks merge from a correctness standpoint.

davydog187:
## Automated review — round 3

Reviewed the current remote tip of `errors/quality-pass` (HEAD `2d4a210`). Re-fetched, built in a detached worktree, and ran the affected suites.

### Verification
- `mix compile --warnings-as-errors` — clean
- `mix test` on `error_gallery_test.exs`, `error_messages_test.exs`, `error_to_map_test.exs`, `for_loop_register_test.exs` — 72 passed, 0 failed

### Correctness
- Render reordering in `format/3` (location-first, no double header, `Enum.join("")` + `String.trim_leading/1`) is consistent. The `Lua runtime error: ` prefix from `Lua.RuntimeException` lands directly before `at <source>:<line>:`, matching the gallery's inline expectations.
- ANSI gating via the `color/2` helper keyed on `IO.ANSI.enabled?/0` is applied at every former unconditional `IO.ANSI.*` site (header removal, location, caret, stack header, omitted-frame line, suggestion). No leak path remains; the `to_map/3` wire path was already escape-free and is untouched.
- `RuntimeError.stringify/1` now matches PUC-Lua: strings verbatim, numbers via `Value.to_string/1`, everything else (table/function/boolean/nil/userdata) as `(error object is a TYPE value)`. Dropping the `stringify(nil)` clause correctly routes `nil` to `(error object is a nil value)`, which is the PUC behavior.
- Suggestion atoms realigned to the kinds the executor actually emits (`:arithmetic_on_non_number`, `:compare_incompatible_types`, `:length_not_integer`, `:bitwise_on_non_integer`, `:for_loop_non_number`); the dead `:arithmetic_type_error` clause is gone. Verified `coerce_for_value/2` raises `:for_loop_non_number`, so the gallery's for-loop case correctly carries a suggestion despite having no location.
- Lua 5.3 semantics spot-checked: `1 < \"x\"` → compare number with string, `1 & \"x\"` → bitwise on string value. Correct.

### Tests
- Gallery cases assert exact equality (`==`) against inline expected strings, so they would fail loudly on the old header/ANSI format — they genuinely pin the new render. Edge cases (no-location for-loop and stack-overflow, error-object table/number/string, assert with/without message) are all covered honestly.

### Findings
- **nit (unfixable without force-push):** the commit `chore(A26): mark plan as review` uses the plan id as the conventional-commit scope; repo convention requires the subsystem (`chore(plan): ...`). The PR title itself is correct (`feat(error): ...`) and the source/test/comment artifacts carry no plan-id leaks. Since rewording would require a force-push (forbidden), flag for awareness only.
- Pre-existing `A18`/`A19` references in `test/lua/error_messages_test.exs:477-478` are from PR #215, not this diff — out of scope here.

Verdict: clean. Correct Lua 5.3 semantics, behavior-pinning golden tests, verified ANSI gating, no source-level convention violations. No blocking findings.

Issue #311, opened 2026-06-01
perf(stdlib): use io_lib.format for string.format float conversion


davydog187:
## Context

`string.format` with many specifiers is **1.85× slower than luerl** (n=1000:
7.42ms vs 4.00ms in `benchmarks/string_format.exs`). The dominant per-call
cost is `format_spec_float/2` at `lib/lua/vm/stdlib/string.ex:454-501`, which
calls `:erlang.float_to_binary` and then post-processes through
`String.split`, `String.pad_trailing`, and string interpolation in
`expand_float/2`.

Luerl (`deps/luerl/src/luerl_lib_string_format.erl:248-250`) delegates the
entire conversion to `io_lib:format/2`, handling precision, sign, and
rounding in a single native Erlang call.

## Change

`lib/lua/vm/stdlib/string.ex:454-501` (`format_spec_float/2`) and likely
`format_spec_scientific/3` at line 503.

- Replace
  `:erlang.float_to_binary([{:decimals, P}, :compact])` + `expand_float`
  with a direct `:io_lib.format(~c\"%.*f\", [precision, val])`.
- Apply the same pattern for `%e` and `%g`.
- Delete `expand_float/2`.

## Compatibility check (required)

Confirm rounding matches C Lua across:
- `test/lua/vm/stdlib/string_test.exs` (all existing format cases)
- Lua 5.3 suite tests touching format/literals

Add unit tests around: precision-0 rounding (sign matters at boundaries),
large precision, `inf`/`nan` formatting.

## Verification

- `mix test` (full suite)
- `mix run benchmarks/string_format.exs`

## Expected

Closes the remainder of the spec-heavy gap. May flip us ahead of luerl on
`many specs`.

Issue #309, opened 2026-06-01
perf(stdlib): parse string.format flags into bitmask, integer specifier


davydog187:
## Context

`string.format` per-specifier paths re-scan the flags binary with
`String.contains?` on every specifier site, and dispatch on the conversion
character as a one-byte binary (e.g. `\"d\"`) rather than an integer (e.g.
`?d`). Luerl precomputes a bitmask and dispatches on `\$d`
(`deps/luerl/src/luerl_lib_string_format.erl:72-156`), which is faster under
BEAM pattern matching.

## Change

`lib/lua/vm/stdlib/string.ex` — `apply_format_spec/2` at line 413 and the
parser that produces `{flags, width, precision, specifier}` tuples.

- Parse flags into a small tuple of booleans (or single integer mask)
  once at parse time.
- Store the conversion char as `?d`/`?f`/… (integer) rather than a binary.

Pure internal refactor — observable output unchanged.

## Verification

- `mix test test/lua/vm/stdlib/string_test.exs`
- Full `mix test`
- `mix run benchmarks/string_format.exs`

## Expected

~10 % of the spec-heavy gap. Stack with T4/T5 for the full close.

PR #316, opened 2026-06-01
perf(stdlib): defer string.format width padding to iolist

### LINKED ISSUE #310: perf(stdlib): defer string.format width padding to iolist
## Context

`string.format` with width-flagged specifiers is **2.03× slower than luerl**
(width-flagged, n=1000: 5.51ms vs 2.71ms in `benchmarks/string_format.exs`).
The literal-heavy fast path from PR #299 is now winning by 2.6×, so the
remaining work is in per-specifier width padding.

`apply_width_flags/3` at `lib/lua/vm/stdlib/string.ex:828` builds
`String.duplicate(pad_char, deficit)` and concatenates with `<>`, producing a
fresh binary per specifier. Luerl returns `[Padding, Bin]` as iolist
(`deps/luerl/src/luerl_lib_string_format.erl:303-319`) and lets the final
`iolist_to_binary` materialise everything at once.

## Change

`lib/lua/vm/stdlib/string.ex:828-859` plus its caller in `format_string/3`.

- `apply_width_flags` returns iolist (`[pad, str]` / `[str, pad]`) instead
  of a binary.
- Thread iolist through `format_string` so the existing accumulator from
  PR #299 absorbs the padding.
- Materialise once at the top level via `IO.iodata_to_binary/1`.

## Verification

- `mix test test/lua/vm/stdlib/string_test.exs`
- Full `mix test`

## Expected

Closes most of the 1.5–2× width-flag gap.

davydog187:
## defer string.format width padding to iolist

Plan: [`.agents/plans/B13-string-format-width-iolist.md`](https://github.com/tv-labs/lua/blob/perf/string-format-width-iolist/.agents/plans/B13-string-format-width-iolist.md)
Closes #310

### Goal

Eliminate the per-specifier binary allocation in `string.format` width padding. `apply_width_flags/3` built `String.duplicate(pad_char, deficit)` and concatenated the padding onto the formatted value with `<>`, producing a fresh binary for every width-flagged specifier. It now returns an **iolist** (`[pad, str]` for right-justify, `[str, pad]` for left-justify) that flows through the iolist accumulator introduced in PR #299, so the padded result is materialised exactly once at the top level via `IO.iodata_to_binary/1`.

This is a **performance-only** change. Output is byte-identical to the previous implementation for every input.

### Success criteria

- [x] `apply_width_flags/3` returns an iolist (`[pad, str]` / `[str, pad]`) instead of a concatenated binary; the no-padding branch still returns `str` unchanged.
- [x] The padded result threads through `format_string/3` / `format_directive/3` with no intermediate `IO.iodata_to_binary` per specifier — materialisation happens only at the `format_string/3` base case. Verified: single call site at `apply_format_spec/2`, which feeds the existing `[acc, str]` append.
- [x] Output byte-identical for left-justify, right-justify, space pad, and zero pad — including the zero-padded-with-sign case (`%05d` of `-7` -> `-0007`). Verified with direct `Lua.eval!` checks and the full property suite.
- [x] Byte-width semantics for multibyte `%s` preserved (`format("%6s", "café")` -> `" café"`, one fill byte). Verified.
- [x] String tests pass: `test/lua/vm/string_test.exs` 152 passed (41 properties, 111 tests). (Note: the path named in the plan, `test/lua/vm/stdlib/string_test.exs`, does not exist; the actual string.format coverage lives in `test/lua/vm/string_test.exs` — recorded in Discoveries.)
- [x] Full `mix test` passes: 2114 passed, 19 skipped, 1 excluded — identical to pre-change counts.
- [x] `mix compile --warnings-as-errors` passes.
- [x] `mix run benchmarks/string_format.exs` width-flagged (n=1000) case improved (see below).

### Benchmark (width-flagged, n=1000)

Focused timing (200 runs after 50-warmup, this machine):

| | ms/call |
|---|---|
| before (main) | ~3.88 |
| after | ~3.39 |

~13% faster on the width-flagged path by removing the per-specifier `<>` copy of `str` + `pad`.

### Changes

```
 .agents/plans/B13-string-format-width-iolist.md | 139 ++++++++++++++++++++++++
 lib/lua/vm/stdlib/string.ex                     |  13 ++-
 2 files changed, 148 insertions(+), 4 deletions(-)
```

### Discoveries

- The plan referenced `test/lua/vm/stdlib/string_test.exs`, which does not exist. The string.format unit + property coverage is in `test/lua/vm/string_test.exs`; that file was run instead (152 passed). No scope change.

### Verification

```
mix format
mix test test/lua/vm/string_test.exs   # 152 passed (41 properties, 111 tests)
mix test                               # 2114 passed, 19 skipped, 1 excluded
```

- Any change to formatting semantics (width/precision/flags, rounding, sign, octal/hex/float output).
- The numeric formatting helpers (`format_spec_integer`, `format_spec_float`, `format_spec_hex`, ...) — they keep returning binaries.
- `format_string/3` / `format_directive/3` parsing, the literal fast path, the `:binary.split/2` chunking from PR #299.
- Any other code in `string.ex`. #309 and #311 also edit this file; this PR stays strictly inside `apply_width_flags/3` and its single call site.
- Codepoint-vs-byte width changes — width stays measured in bytes.

### DIFF
--- a/.agents/plans/B13-string-format-width-iolist.md
+++ b/.agents/plans/B13-string-format-width-iolist.md
@@ -2,10 +2,10 @@
 id: B13
 title: defer string.format width padding to iolist
 issue: 310
-pr: null
+pr: 316
 branch: perf/string-format-width-iolist
 base: main
-status: in-progress
+status: review
 direction: B
 ---
 
@@ -137,3 +137,27 @@ demonstrate the improvement (target: close most of the ~2.03× luerl gap).
   base case.
 - **No expected suite-count change.** This is perf-only; `mix test`
   pass/fail counts should be identical before and after.
+
+## What changed
+
+- `lib/lua/vm/stdlib/string.ex`: `apply_width_flags/3` now returns an
+  iolist for the padded cases — `[str, pad]` (left-justify),
+  `["-", pad, binary_part(...)]` (zero-pad with leading sign), and
+  `[pad, str]` (right-justify default). The no-padding branch still
+  returns the bare `str`. The outer `<>` concatenation that copied
+  `str` + `pad` into a new binary per specifier is gone; the result
+  threads through the existing `[acc, str]` append in
+  `format_directive/3` and is materialised once at the
+  `format_string/3` base case via `IO.iodata_to_binary/1`. Single call
+  site (`apply_format_spec/2`) needed no change.
+- `.agents/plans/B13-string-format-width-iolist.md`: lifecycle commits.
+
+Verification: `mix test` 2114 passed / 19 skipped / 1 excluded (identical
+to pre-change). String coverage `test/lua/vm/string_test.exs` 152 passed.
+Output byte-identical across left/right/space/zero padding, zero-pad-with-sign
+(`%05d` of `-7` -> `-0007`), and multibyte `%s` (`%6s` of `"café"` -> `" café"`).
+Width-flagged benchmark (n=1000): ~3.88 ms/call -> ~3.39 ms/call (~13% faster).
+
+Discovery: the plan named `test/lua/vm/stdlib/string_test.exs`, which does
+not exist; string.format coverage lives in `test/lua/vm/string_test.exs`.
+No scope change.
--- a/.agents/plans/B13-string-format-width-iolist.md
+++ b/.agents/plans/B13-string-format-width-iolist.md
@@ -1,0 +1,139 @@
+---
+id: B13
+title: defer string.format width padding to iolist
+issue: 310
+pr: null
+branch: perf/string-format-width-iolist
+base: main
+status: in-progress
+direction: B
+---
+
+# B13 — defer string.format width padding to iolist
+
+## Goal
+
+Eliminate the per-specifier binary allocation in `string.format` width
+padding. `apply_width_flags/3` currently builds
+`String.duplicate(pad_char, deficit)` and concatenates the padding onto
+the formatted value with `<>`, producing a fresh binary for every
+width-flagged specifier. Make `apply_width_flags/3` return an **iolist**
+(`[pad, str]` for right-justify, `[str, pad]` for left-justify) and let it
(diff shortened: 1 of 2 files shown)


davydog187 (commented):
## Automated code review — verdict: clean (perf-only, correct)

Reviewed the diff against `main`. This is a sound performance-only change; output is byte-identical and the data flow is iolist-safe end to end.

**Correctness (the one real risk: does anything downstream treat the result as a binary?) — no.**
- `apply_width_flags/3` measures `byte_size(str)`, `binary_part(str, …)`, and `String.starts_with?(str, "-")` on the **input `raw` binary**, then returns an iolist. No binary operation is ever applied to the iolist it produces.
- Its sole consumer is `apply_format_spec/2` → `format_directive/3`, which appends `[acc, str]`. An iolist `str` is a valid iolist element, so it threads through untouched and is materialised exactly once at the `format_string/3` base case (`IO.iodata_to_binary/1`). No intermediate flatten, no per-specifier binary allocation — which is the whole point.

**Edge cases verified:**
- Zero-pad-with-sign `["-", pad, binary_part(str, 1, byte_size(str) - 1)]` is byte-for-byte equal to the old `"-" <> pad <> binary_part(...)` — sign stays leftmost (`%05d` of `-7` → `-0007`).
- `0`+`-` flag combo: `pad_char` correctly falls back to space (printf has `-` override `0`); left-justify path `[str, pad]`.
- Multibyte `%s` width stays byte-measured (`%6s` of `"café"` → `" café"`).

**Conventions:** scope is `stdlib` (not the plan id); no plan-id leakage in source; comment explains the iolist rationale well.

**Tests/bench:** full suite byte-identical (2114 passed / 19 skipped, unchanged); width-flagged n=1000 ~3.88ms → ~3.39ms (~13% faster).

No blocking or actionable findings. Note for the merge train: #317 and #319 also edit `lib/lua/vm/stdlib/string.ex`, so whichever of these merges later will need a trivial rebase.

PR #318, opened 2026-06-01
perf(stdlib): batch write-back in table.sort plain fast path

### LINKED ISSUE #307: perf(stdlib): batch write-back in table.sort plain fast path
## Context

`table.sort` on plain (metatable-less) tables is **1.44× slower than luerl**
at n=100 (46.1µs vs 31.9µs in `benchmarks/table_ops.exs`).

The plain-table sort fast path (PR #299) reads + sorts efficiently, but the
**write-back loop** still calls `Table.put/3` per index, which rebuilds the
seven-field `%Table{}` struct on every iteration and updates `state.tables`.
Luerl's `array:from_list` writes the sorted slice back in a single step
(`deps/luerl/src/luerl_lib_table.erl:363-373`).

## Change

`lib/lua/vm/stdlib/table.ex:312-326` — replace:

\`\`\`elixir
updated =
  sorted
  |> Enum.with_index(1)
  |> Enum.reduce(table, fn {val, idx}, tbl -> Table.put(tbl, idx, val) end)
\`\`\`

with a single map build and one `Table.replace_data/2` call. `replace_data`
already exists. Sort doesn't touch order/dead-key state, so the simpler
replacement is safe.

## Verification

- `mix test` (full suite stays green)
- `mix test test/lua/vm/stdlib/table_test.exs`
- `mix run benchmarks/table_ops.exs` — record new ratio in the PR body

## Expected

Closes ~30 % of the 1.44× sort gap. Target: ≤ 1.10× luerl.

davydog187:
## perf(stdlib): batch write-back in table.sort plain fast path

Plan: [B10](.agents/plans/B10-table-sort-batch-writeback.md)
Closes #307

## Goal

`table.sort` on plain (metatable-less) tables wrote the sorted slice back
with an `Enum.reduce` that called `Table.put/3` once per index, rebuilding
the seven-field `%Table{}` struct and walking `order`/`dead` bookkeeping on
every iteration — O(n) struct rebuilds. Sort rewrites every integer key
`1..len` and never depends on the prior `order`/`dead` state, so the
write-back is replaced with a single map build plus one
`Table.replace_data/2` call.

## Success criteria

- [x] `mix format` clean and `mix compile --warnings-as-errors` passes (no
      warnings from the `lua` app).
- [x] `mix test` full suite green: `2114 passed, 19 skipped, 1 excluded`.
- [x] `mix test test/lua/vm/stdlib/table_test.exs` passes: `60 passed
      (6 properties, 54 tests)` — sorted result, custom comparator,
      comparator-mutates-table, and `len <= 1` handling unchanged.
- [x] Benchmark recorded below. The change is a consistent ~12% absolute
      speedup on the plain-sort path. On this M4 hardware luerl is unusually
      fast (~21µs vs the issue's recorded 31.9µs), so the luerl ratio lands
      at ~1.47x rather than the ≤1.10x projection — see "Benchmark" note.
- [x] No behavior change: the write-back produces the same final table
      contents and `order`/`dead` state as a fresh sequence of `Table.put/3`
      writes over keys `1..len` (sort assigns every key `1..len` live;
      `replace_data/2` rebuilds `order` from `Map.keys/1` and clears `dead`).

## Benchmark

`MIX_ENV=benchmark mix run benchmarks/table_ops.exs` — Table Sort, n=100
(mode: quick), `lua (chunk)` vs `luerl`, same M4 machine:

| build | lua (chunk) | luerl | ratio |
|-------|-------------|-------|-------|
| main (per-index `Table.put/3`) | 34.03 µs | 21.76 µs | 1.56x |
| this PR (`replace_data/2`)     | 30.06 µs | 20.45 µs | 1.47x |

The change is a consistent ~12% reduction in the plain-sort path
(34.0µs → 30.0µs across repeated passes) and a strict reduction in struct
rebuilds. The absolute ≤1.10x target from the issue was set against a
recorded luerl baseline of 31.9µs; on this M4 luerl benches at ~21µs, so the
ratio target is not reachable by this single optimization alone. Per the
plan's Risks section, this is shipped as a genuine improvement with the
actual ratio recorded; closing the remaining gap is follow-up work, not an
expansion of this PR's scope.

## Changes

```
 .agents/plans/B10-table-sort-batch-writeback.md | 136 ++++++++++++++++++++++++
 lib/lua/vm/stdlib/table.ex                      |  12 ++-
 2 files changed, 145 insertions(+), 3 deletions(-)
```

## Out of scope

- The metatable-backed sort path (`sort_via_metamethods/4`).
- The comparator/sort algorithm (`sort_values/3`) and the read loop.
- Any change to `Table.replace_data/2`, `Table.put/3`, or `%Table{}`.
- `table.insert`/`table.remove`/other table ops; array-part storage rework.

### DIFF
--- a/.agents/plans/B10-table-sort-batch-writeback.md
+++ b/.agents/plans/B10-table-sort-batch-writeback.md
@@ -1,0 +1,136 @@
+---
+id: B10
+title: Batch write-back in table.sort plain fast path
+issue: 307
+pr: null
+branch: perf/table-sort-batch-writeback
+base: main
+status: in-progress
+direction: B
+---
+
+## Goal
+
+`table.sort` on plain (metatable-less) tables is ~1.44x slower than luerl
+at n=100 (46.1µs vs 31.9µs in `benchmarks/table_ops.exs`). The plain-table
+fast path (PR #299) reads and sorts efficiently, but the write-back loop in
+`lib/lua/vm/stdlib/table.ex` (~312-326) still calls `Table.put/3` once per
+index inside an `Enum.reduce`. Each call rebuilds the seven-field
+`%Table{}` struct and walks the `order`/`dead` insertion-order bookkeeping,
+so writing back an n-element slice is O(n) struct rebuilds.
+
+Replace the per-index reduce with a single map build plus one
+`Table.replace_data/2` call. `replace_data/2` already exists
+(`lib/lua/vm/table.ex:80`): it sets `data`, rebuilds `order` from
+`Map.keys(data)`, and clears `dead` in one struct update. Because sort
+rewrites every integer key `1..len` and never depends on the prior
+`order`/`dead` state, the wholesale replacement is observably identical to
+the per-index writes.
+
+This is a perf-only change. Observable behavior — sorted result, error
+behavior, metatable path, comparator-mutation re-fetch — is unchanged.
+
+Target: bring the plain-sort path to ≤ 1.10x luerl at n=100.
+
+## Out of scope
+
+- The metatable-backed sort path (`sort_via_metamethods/4`). It must keep
+  reading via `__index` and writing via `__newindex` so metamethod
+  observation order matches the reference impl; it is untouched here.
+- The comparator/sort algorithm itself (`sort_values/3`) and the read loop
+  (`Enum.map(1..len//1, ...)`). Only the write-back step changes.
+- Any change to `Table.replace_data/2`, `Table.put/3`, or the `%Table{}`
+  struct shape. This plan only changes the *caller* in the stdlib.
+- Optimizing `table.insert`, `table.remove`, or other table stdlib ops.
+- Array-part / threshold-promotion storage rework (the deferred B7 idea).
+
+## Success criteria
+
+- [ ] `mix format` clean and `mix compile --warnings-as-errors` passes.
+- [ ] `mix test` full suite green with no regressions.
+- [ ] `mix test test/lua/vm/stdlib/table_test.exs` passes — sort behavior
+      (sorted result, custom comparator, comparator that mutates the table,
+      stable handling of `len <= 1`) unchanged.
+- [ ] `mix run benchmarks/table_ops.exs` shows the plain `table.sort` path
+      at ≤ 1.10x luerl at n=100 (down from ~1.44x). New ratio recorded in
+      the PR body.
+- [ ] No behavior change: the write-back produces the same final table
+      contents and `order`/`dead` state as a fresh sequence of `Table.put/3`
+      writes over keys `1..len`.
+
+## Implementation notes
+
+Single file: `lib/lua/vm/stdlib/table.ex`, in `sort_plain/5` (~312-326).
+
+Replace the per-index reduce:
+
+```elixir
+updated =
+  sorted
+  |> Enum.with_index(1)
+  |> Enum.reduce(table, fn {val, idx}, tbl -> Table.put(tbl, idx, val) end)
+
+{[], %{state | tables: Map.put(state.tables, id, updated)}}
+```
+
+with a single map build and one `Table.replace_data/2` call, e.g.:
+
+```elixir
+data =
+  sorted
+  |> Enum.with_index(1)
+  |> Map.new(fn {val, idx} -> {idx, val} end)
+
+updated = Table.replace_data(table, data)
+
+{[], %{state | tables: Map.put(state.tables, id, updated)}}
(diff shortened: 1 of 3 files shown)


davydog187 (commented):
**Verdict: clean-with-nits** — the optimization is correct and behavior-preserving; the only issues are a convention violation on commit scope and stale prose that describes a different (and unsafe) implementation than the one that shipped.

_Automated review (skeptical senior-Elixir pass). I read the diff, `Lua.VM.Table` (`put/3`, `replace_data/2`), `Value.sequence_length/1`, and the executor gating; CI is green on all five checks._

### What I verified (correctness — passes)

- **The write-back is observably identical to N× `Table.put/3`.** `table_length` is `Value.sequence_length/1`, which returns the largest N where keys `1..N` are all present (`lib/lua/vm/value.ex:166`). So every key `1..len` is a *live* entry in `data`. The old `Enum.reduce(... Table.put/3 ...)` therefore always hit the "update existing live key" branch (`lib/lua/vm/table.ex:125-127`) — `Map.put(data, k, v)` only, leaving `order`/`order_tail`/`dead` untouched. The new `%{table | data: Map.merge(table.data, sorted_slice)}` (`lib/lua/vm/stdlib/table.ex:325`) performs the same value overwrites and leaves `order`/`order_tail`/`dead` literally untouched. Equivalent. ✓
- **Non-array keys survive.** `Map.merge` keeps non-overlapping keys, so `t.name`, sparse `t[100]`, etc. are preserved — matching the old `put/3` path which never touched them. The new regression test (`table_test.exs:189-204`, `"table.sort preserves non-array keys"`) pins exactly this. ✓
- **Metatable gating is intact.** `sort_plain` is only reached when `table.metatable == nil` (`executor`/`table_sort` dispatch at `lib/lua/vm/stdlib/table.ex:284`); `__index`/`__newindex` tables still take `sort_via_metamethods`. Unchanged by this PR. ✓
- **Scope is tight.** Only the `sort_plain/5` write-back + its comment, one regression test, and the plan file. `sort_via_metamethods`, `sort_values`, and the `%Table{}` struct are untouched. ✓

### Findings

- **[minor] PR body, plan, and commit `a8215532` describe a different — and *unsafe* — implementation than what shipped.** All the prose (PR "Goal"/"Success criteria", `B10-*.md`, the `perf(stdlib)` commit body) says the write-back uses `Table.replace_data/2`. The shipped code uses `Map.merge` instead, and that difference is load-bearing: `replace_data/2` rebuilds `order` from `Map.keys(data)` and clears `dead` (`lib/lua/vm/table.ex:80-82`), which would have **reordered non-array keys in `pairs` iteration and dropped dead-key tracking** — i.e. it would *not* have been behavior-preserving for a table with string/sparse keys. The commit history shows this was caught: `f00466cc` ("address review feedback") switched `replace_data` → `Map.merge` and added the regression test. Good catch in review — but please update the PR body / success-criteria bullet (and ideally the plan's "Implementation notes") so they describe `Map.merge`, not `replace_data/2`. As written, the "No behavior change" bullet justifies the wrong implementation.

- **[minor] Convention: housekeeping commits use the plan id as the subject scope.** `chore(B10): start plan` and `chore(B10): mark plan as review` put `B10` in the `scope` position. Per CLAUDE.md and `ship-a-plan/SKILL.md`, commit subjects must use the affected subsystem as scope, **never the plan id** (the plan id belongs in the commit *body* as `Plan: B10`, which the main commit does correctly). The feature commit `perf(stdlib): ...` is correct; only the `chore(B10)` ones violate. Not a code blocker, and squash-merge would erase them, but flagging since it's an explicit repo rule.

- **[nit] `Table.replace_data/2` remains defined but unused.** This is pre-existing (it lives on `origin/main`, originally added in A7a "for table.sort" but never actually wired up there), so it's *not* introduced by this PR. Worth noting only because this PR was the natural place to either use it or remove it, and instead it cements the dead-code status. Optional follow-up, out of scope here.

### Non-issues confirmed

- No `Co-Authored-By` AI trailer. ✓
- No plan id leaked into any source or test file (only in `.agents/plans/B10-*.md`, which is correct). ✓
- CI green: build (1.19 + 1.20-rc), Dialyzer, Lua 5.3 suite, website. No red tests. ✓
- `len <= 1` early-return and the comparator-mutation re-fetch are preserved unchanged. ✓

The benchmark missing its ≤1.10× target is adequately explained (luerl baseline shifted on the M4 hardware); the change is a real, strict reduction in struct rebuilds. No objection to shipping on perf grounds.

PR #321, opened 2026-06-01
perf(vm): batch :set_list writes through Table.put_many/2

### LINKED ISSUE #308: perf(vm): batch insert for :setlist instead of per-key Table.put
## Context

Table construction is **1.14× slower than luerl** at n=100 (25.5µs vs 22.3µs
in `benchmarks/table_ops.exs`).

`:setlist` walks one key at a time through `Table.put/3`, paying the
seven-field `%Table{}` struct rebuild on every slot. Luerl's `:array:set`
amortises this to a single mutation.

## Change

- `lib/lua/vm/dispatcher.ex:1412-1415` — `set_list_into_table/6`: accumulate
  `{idx, val}` pairs and call `Table.replace_data/2` once at the end of the
  run instead of `Table.put/3` per slot.
- Mirror the change in the executor's `:setlist` handler (`lib/lua/vm/executor.ex`).

Helper `Table.replace_data/2` already exists in `lib/lua/vm/table.ex`.

## Verification

- `mix test` (full suite stays green)
- `mix test test/lua/vm/stdlib/table_test.exs`
- `mix run benchmarks/table_ops.exs`

## Expected

Closes most of the 1.14× build gap. Target: ≤ 1.05× luerl.

davydog187:
## batch insert for `:set_list` instead of per-key `Table.put`

Plan: [`.agents/plans/B11-setlist-batch-insert.md`](https://github.com/tv-labs/lua/blob/perf/setlist-batch-insert/.agents/plans/B11-setlist-batch-insert.md)
Closes #308

### Goal

Make `:set_list` (table-constructor backfill) accumulate its `{index, value}` pairs and apply them to the `%Lua.VM.Table{}` in a single mutation, instead of walking one key at a time through `Table.put/3` and paying the seven-field struct rebuild on every slot. Performance-only: stored values, iteration order, dead-key (nil-clear) semantics, and the public `Lua.*` API are unchanged.

### Approach

- Added `Lua.VM.Table.put_many/2`: folds an ordered list of `{key, value}` pairs into the table, making the **exact same** `order`/`order_tail`/`dead` decisions as repeated `put/3` (new keys append, live keys keep position, dead keys revive to the end, `nil` clears into `dead`), but rebuilding the struct only once. Deliberately **not** routed through the existing `replace_data/2`, which would clear `dead` and rebuild `order` from `Map.keys/1` — that is not behavior-identical.
- `dispatcher.ex` `set_list_into_table/6` and **both** `executor.ex` `:set_list` arms (integer-count incl. its `count == 0` multi-return branch, and the `{:multi, init_count}` form) collect their pairs in one register walk and call the shared `put_many/2` once. The two executors stay mirror-parallel via the same helper.

### Success criteria

- [x] `mix test` — full suite green, no regressions (2114 → 2119 passing; +5 from the new constructor tests, 0 failing, 19 skipped).
- [x] `mix test test/lua/vm/stdlib/table_test.exs` passes (60 → 65).
- [x] Iteration order and dead-key semantics unchanged — new tests in `table_test.exs` assert constructor `pairs`/`ipairs` order, `ipairs` stop-at-hole, dead-key revival moving to the end (`1,2,4,5,3`), overwrite keeping position, and empty constructor.
- [x] `mix format` clean.
- [x] luerl ratio at n=100 and n=1000 recorded below.

### Benchmark (table build, `benchmarks/table_ops.exs`, `LUA_BENCH_MODE=full`)

The benchee n=100 microbenchmark is extremely noisy on this machine (deviation ±20–195%; the luerl vs lua-chunk ordering flips run to run), so the recorded benchee ratios are reported alongside a tight back-to-back A/B that isolates the `:set_list` path.

**benchee medians, this build vs luerl:**

| n | luerl (median) | lua (chunk, median) | ratio |
|---|---|---|---|
| 100 | 17.25 µs | 18.17 µs | **1.05×** |
| 1000 | 157.0 µs | 220.4 µs | 1.40× |

Baseline (pre-change) on the same machine measured n=100 build at ~1.04× luerl (median), i.e. the gap the plan cited as 1.14× did not reproduce here — the build was already ~1.0–1.05× luerl, so there was little headroom for the benchee number to move, and the benchee n=100 ratio is within noise of baseline. No regression at n=1000.

**Tight A/B (200k iterations, eval of a 100-element literal constructor; full eval overhead included, so it understates the per-opcode delta), reproducible across runs:**

| build | µs/op |
|---|---|
| baseline (per-slot `put/3`) | 43.7 – 44.4 |
| after (`put_many/2`) | 41.2 – 41.8 |

~5% faster on the set_list-heavy path, stable across repeated runs. Behavior-identical and modestly faster — not a neutral change.

### Changes
```
 lib/lua/vm/dispatcher.ex          | 22 +++++++++----
 lib/lua/vm/executor.ex            | 44 +++++++++++--------------
 lib/lua/vm/table.ex               | 64 ++++++++++++++++++++++++++++++++++++
 test/lua/vm/stdlib/table_test.exs | 68 +++++++++++++++++++++++++++++++++++++++
```

### Verification
```
mix format
mix test
# Result: 2119 passed (60 doctests, 51 properties, 2008 tests), 19 skipped, 1 excluded
mix test test/lua/vm/stdlib/table_test.exs
# Result: 65 passed (6 properties, 59 tests)
PKG_CONFIG_PATH=... MIX_ENV=benchmark LUA_BENCH_MODE=full mix run benchmarks/table_ops.exs
```

- The B7-style array/hash split for `Lua.VM.Table` (deferred).
- Changes to `Table.put/3`, `Table.put_data/3`, or the `order`/`order_tail`/`dead` invariants themselves.
- Other hot dispatch paths (`set_table`, `set_field`, `rawset`).
- Encoder/compiler changes that emit `:set_list`.

### DIFF
--- a/.agents/plans/B11-setlist-batch-insert.md
+++ b/.agents/plans/B11-setlist-batch-insert.md
@@ -2,10 +2,10 @@
 id: B11
 title: batch insert for :setlist instead of per-key Table.put
 issue: 308
-pr: null
+pr: 321
 branch: perf/setlist-batch-insert
 base: main
-status: in-progress
+status: review
 direction: B
 ---
 
@@ -140,3 +140,46 @@ regressed at `n=1000`.
   vararg/multi-return call. Ensure the batch path handles `total == 0`
   (empty) and the multi-return slot offsets identically to the current
   reduce.
+
+## What changed
+
+Shipped as PR #321.
+
+- `lib/lua/vm/table.ex` — added `put_many/2` (plus accumulator-threaded
+  `insert_acc`/`delete_acc` mirrors of `insert/3`/`delete/2`). It folds an
+  ordered `{key, value}` list into the table making the same
+  `order`/`order_tail`/`dead` decisions as repeated `put/3`, but rebuilds
+  the `%Table{}` struct once. Deliberately not routed through
+  `replace_data/2` (which clears `dead` and rebuilds `order` from
+  `Map.keys/1` — not behavior-identical).
+- `lib/lua/vm/dispatcher.ex` — `set_list_into_table/6` now collects the
+  run's pairs via a new `set_list_pairs/7` register walk and applies them
+  with a single `put_many/2`.
+- `lib/lua/vm/executor.ex` — both `:set_list` arms (the `{:multi, _}` form
+  and the integer-count form, including its `count == 0` multi-return
+  branch) route through a shared `set_list_pairs/4` helper + `put_many/2`.
+  `total == 0` yields `[]` and leaves the table untouched.
+- `test/lua/vm/stdlib/table_test.exs` — added a `table constructor
+  backfill` describe block (5 tests) pinning constructor `pairs`/`ipairs`
+  order, `ipairs` stop-at-hole, dead-key revival ordering (`1,2,4,5,3`),
+  overwrite position stability, and the empty constructor.
+
+Suite: 2114 → 2119 passing (+5 new tests), 0 failing, 19 skipped.
+
+Numbers: benchee table-build n=100 ratio ~1.05× luerl (median; n=100 is
+very noisy at ±20–195% deviation), no regression at n=1000. A tight
+back-to-back A/B (200k iterations over a 100-element literal constructor)
+isolates a stable ~5% improvement on the `:set_list` path (baseline
+~43.7–44.4 µs/op → ~41.2–41.8 µs/op after).
+
+## Discoveries
+
+- The plan's cited 1.14× luerl baseline at n=100 did not reproduce on the
+  benchmarking machine; the table-build workload was already ~1.0–1.05×
+  luerl there. The benchee n=100 microbenchmark is too noisy to show the
+  change's effect (relative ordering flips run to run), so the win was
+  confirmed via a tight isolated A/B loop instead. No scope change.
+- `benchmarks/setup_luaport.sh` is required to build the optional
+  `luaport` (C Lua 5.4) dep under `MIX_ENV=benchmark`; without it the
+  whole benchmark env fails to compile (`lua.h` not found). Not changed by
+  this plan.
--- a/.agents/plans/B11-setlist-batch-insert.md
+++ b/.agents/plans/B11-setlist-batch-insert.md
@@ -1,0 +1,142 @@
+---
+id: B11
+title: batch insert for :setlist instead of per-key Table.put
+issue: 308
+pr: null
+branch: perf/setlist-batch-insert
+base: main
+status: in-progress
+direction: B
+---
+
+# B11 — batch insert for `:setlist` instead of per-key `Table.put`
+
+## Goal
+
+Make `:set_list` (table constructor backfill) accumulate its
+`{index, value}` pairs and apply them to the `%Lua.VM.Table{}` in a
+single mutation, instead of walking one key at a time through
+`Table.put/3` and paying the seven-field struct rebuild on every slot.
+
+Table construction is currently **1.14× slower than luerl** at `n=100`
+(25.5µs vs 22.3µs in `benchmarks/table_ops.exs`). Luerl amortises the
+same work to a single `:array:set` mutation. Batching the inserts should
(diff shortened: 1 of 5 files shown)


davydog187 (commented):
**Verdict: clean-with-nits** — behavior-identical batch refactor; one convention nit on commit scope.

> _Automated review_ (skeptical pass focused on `put_many/2` correctness, nil-delete semantics, and executor/dispatcher parity).

### nil-delete semantics — CORRECT
`Table.put_many/2` (`lib/lua/vm/table.ex:181-201`) dispatches per pair on the value exactly as `put/3` does: `nil -> delete_acc`, `_ -> insert_acc`. `delete_acc` (`table.ex:225-231`) and `insert_acc` (`table.ex:206-222`) are line-for-line transcriptions of `delete/2` (`table.ex:136-147`) and `insert/3` (`table.ex:107-134`), operating on the bare `{data, order, order_tail, dead}` tuple instead of the struct. Revive branch returns `order_tail: [key]`, dead/no-op branches return state unchanged — all matching the struct version. So an assigned `nil` clears a live key into `dead` and is a no-op on an absent key, identical to a `put/3` fold. `metatable` is untouched and preserved by the single `%{table | ...}` rebuild. **Folding bare fields is observationally identical to chaining `put/3`** because each `put/3` writes exactly the four fields the next read consumes.

### executor/dispatcher parity — CORRECT (both paths converted)
- Dispatcher `set_list_into_table/6` (`dispatcher.ex`) now calls `set_list_pairs/7` then one `put_many/2`. With caller `i=0`, walking `j` from `count-1`→0 and prepending yields ascending keys `offset+1..offset+count` with values `element(start+1..start+count)` — identical to the old `i=0..count-1` recursion.
- Executor: **both** arms converted — the `{:multi, init_count}` arm (uses pre-existing `total = init_count + multi_return_count`) and the integer-count arm (new `total = if count == 0, do: multi_return_count, else: count`). Both route through `set_list_pairs/4` (`executor.ex`), which walks `k` from `total-1`→0: key `offset+k+1`, value `elem(regs, start+k)`. This matches the old `total>0` reduce and, for the non-zero integer branch, the old `1..count` reduce (`offset+i` / `elem(start+i-1)` re-indexes to the same pairs). `total == 0` → `[]` → `put_many(table, [])` no-ops via the `[]` clause (`table.ex:179`), matching the old `else table` branches. The two sites stay mirror-parallel via the shared `put_many/2`. **No parity gap.**

### [nit] commit scope uses plan id `B11`
`chore(B11): start plan` and `chore(B11): mark plan as review` (commit subjects) use the plan id as the conventional-commit scope. `CLAUDE.md` / `.agents/skills/ship-a-plan/SKILL.md` require the affected subsystem as scope, never the plan id. The PR title (`perf(vm): ...`) is correct, and `Plan: B11` correctly lives in a commit *body* (explicitly allowed), and no `B11` leaks into source/tests — only the two `chore(B11)` subjects deviate. Non-blocking but worth squashing/relabeling before merge.

### [nit] no explicit test for a non-zero `offset` set_list run
Tests cover constructor pairs/ipairs order, ipairs-stop-at-hole, dead-key revival (`1,2,4,5,3`), overwrite position stability, and the empty constructor — a good set, and the revival test is the one that would catch a fold mis-thread. There's no test exercising a `:set_list` run with a non-1 base `offset` (positional+keyed constructor mix) or a multi-run / large (>FIELDS_PER_FLUSH) constructor. Low risk: the `offset + k + 1` arithmetic is carried through unchanged from the original, so it can't drift in this refactor — but a constructor like `{1,2,3, [10]=x, 4,5}` would pin the offset path explicitly. Optional.

### scope — clean
Diff touches exactly the four in-scope files plus the plan doc: `dispatcher.ex`, `executor.ex`, `table.ex`, `table_test.exs`, `.agents/plans/B11-*.md`. No stray edits to `put/3`, `replace_data/2`, or other dispatch paths. The PR body is also commendably honest that the benchee n=100 win is within noise and leans on a tight A/B instead — not overclaimed.

**Recommendation:** merge after fixing the `chore(B11)` commit-scope nit (and optionally adding an offset test).

PR #325, opened 2026-06-02
perf(vm): document why depth-1 upvalues cannot be inlined in the closure tuple

### LINKED ISSUE #313: perf(vm): inline depth-1 upvalues in closure tuple
## Context

The closures benchmark is **1.22× slower than luerl** (693µs vs 568µs in
`benchmarks/closures.exs`).

Every `:get_upvalue` does a `Map.get(state.upvalue_cells, cell_ref)` and
every `:set_upvalue` rebuilds the state struct via
`%{state | upvalue_cells: Map.put(...)}` (`lib/lua/vm/executor.ex:666-680`).
Luerl (`deps/luerl/src/luerl_emul.erl:208-213`) has a depth-1/2 fast path
that short-circuits the map lookup for the common case.

The closures bench reads its captured `count` on every `counter()` call, so
this overhead dominates.

## Change

`lib/lua/vm/executor.ex:666-680` (`:get_upvalue`, `:set_upvalue`) plus the
open-upvalue siblings near line 688.

Where possible, keep upvalue values inline in the closure's `upvalues`
tuple. Fall back to `state.upvalue_cells` only when the cell genuinely
escapes — closure-over-closure or aliased through `:set_open_upvalue`.

Match luerl's approach: depth-1 access is direct tuple read; deeper cases
pay the map cost.

## Verification

- `mix test` (full suite — upvalue semantics are critical)
- `mix test test/lua/vm/closures_test.exs`
- Lua 5.3 suite (closures.lua and friends)
- `mix run benchmarks/closures.exs`
- `mix run benchmarks/fibonacci.exs` (no regression)

## Expected

Most of the 1.22× closures gap.

davydog187:
## Summary

Investigated **#313** — inlining depth-1 upvalue values directly into the
closure `upvalues` tuple so `:get_upvalue` / `:set_upvalue` can read a value
without touching `state.upvalue_cells`.

**Outcome: the proposed inlining is not feasible within tight scope, and the
map access it targets is not the actual bottleneck.** This PR documents the
constraint in the executor instead of churning the correctness-critical hot
path. The diff is comment-only — no behavioral change.

> [!WARNING]
> This is **not** a perf win. It is the honest result of the investigation.
> The maintainer should decide whether to close #313 as not-actionable-as-
> specified, or to pursue the deeper closure-value rewrite described below.

## Why inlining doesn't work

The `upvalues` tuple is captured **by value** into the immutable closure
term. The upvalue value, however, is **mutable shared state**:

- A `set_upvalue` has no reference back to the closure term, so an inlined
  value could never persist across calls of the same closure, nor become
  visible to a sibling closure that captured the same parent local.
- The cell ref is a cross-module contract: the compiled dispatcher
  (`Lua.VM.Dispatcher`), `_ENV` rebinding (`Stdlib`), and
  `debug.getupvalue`/`setupvalue` all read and write `state.upvalue_cells`
  through it. `debug.setupvalue` can mutate *any* upvalue at runtime, so even
  a read-only-inlining variant is unsound.

The shared, call-to-call-persistent backing store **is** `state.upvalue_cells`,
threaded through `State` — this is the same heap-cell model luerl uses
(`luerl_heap:get_env_var`). luerl's depth-1/2 "fast path" only skips the
`lists:nth/2` env-**frame** walk; our `elem/2` tuple read already avoids that
walk, and luerl still pays the same heap cost.

## Why the map access isn't the bottleneck

Isolated micro-benchmark on a 50-entry ref-keyed map (this machine):

| op | ips |
|----|-----|
| `Map.get` | 55.9 K |
| `:erlang.map_get` | 59.1 K (~5% faster) |
| struct update via `Map.put` | 40.3 K |
| struct update via `:maps.put` | 40.1 K (no better) |

~5% on one opcode is noise inside the full VM loop, and the closures workload
is dominated by per-call closure setup, not the upvalue cell access.

## Benchmarks (back-to-back, same machine + load)

Captured `main` and this branch back-to-back to control for luerl's large
run-to-run variance.

**closures.exs (`lua (chunk)`), MIX_ENV=benchmark:**

| | avg | ips |
|---|-----|-----|
| before (main) | 493.97 µs | 2.02 K |
| after (branch) | 501.46 µs | 1.99 K |
| delta | +1.5% (within ±5% deviation — flat) | |

luerl in the same runs ranged 416–567 µs; we are at/near parity, not the
1.22× slower the issue measured on a different machine.

**fibonacci.exs (`lua (chunk)`) — no upvalue opcodes, regression guard:**

| | avg |
|---|-----|
| before (main) | 843.18 ms |
| after (branch) | 842.77–848.87 ms |
| delta | flat (noise) |

## Correctness

Comment-only change; behavior is identical. Full validation run:

- `mix format` — clean
- `mix test` — **2153 passed, 19 skipped, 0 failures** (incl. the Lua 5.3
  official suite and `test/lua/vm/upvalue_test.exs` aliasing/block-exit cases)
- `mix test test/lua53_suite_test.exs` — 17 passed, 12 skipped (closure.lua &
  friends)
- No `mix credo` in this project

## A real win would require

Re-architecting closures so a closure carries its own mutable upvalue boxes
(e.g. per-closure cell storage) instead of indexing a single shared
`state.upvalue_cells` map — a cross-cutting change to the interpreter,
dispatcher, bytecode, `_ENV`, and the `debug` upvalue API. That is well
beyond a single-opcode perf PR and would need its own design.

Closes #313

### DIFF
--- a/lib/lua/vm/executor.ex
+++ b/lib/lua/vm/executor.ex
@@ -662,7 +662,23 @@ defmodule Lua.VM.Executor do
     do_execute(rest, regs, upvalues, proto, state, cont, frames, line)
   end
 
-  # ── get_upvalue ────────────────────────────────────────────────────────────
+  # ── get_upvalue / set_upvalue ──────────────────────────────────────────────
+  #
+  # The closure's `upvalues` tuple holds the cell ref directly, so resolving a
+  # depth-N upvalue is already a constant-time `elem/2` read — there is no
+  # frame-list walk to short-circuit the way luerl's `get_env_var_1/4`
+  # (`luerl_emul.erl`) skips `lists:nth/2` for depth 1 and 2.
+  #
+  # The upvalue *value* cannot be inlined into the tuple. The tuple is captured
+  # by value into the (immutable) closure term, while the value is mutable
+  # shared state: a later `set_upvalue` (or `set_open_upvalue`, or
+  # `debug.setupvalue`) must be visible to the next call of this closure and to
+  # any sibling closure that captured the same parent local. That shared,
+  # call-to-call-persistent store is `state.upvalue_cells`, keyed by the cell
+  # ref and threaded through `State`; the compiled dispatcher (`Lua.VM.Dispatcher`),
+  # `_ENV` rebinding, and `debug.getupvalue`/`setupvalue` all read and write the
+  # same cell. `Map.get/2` keeps the nil-for-dangling-cell semantics the
+  # dispatcher mirrors.
 
   defp do_execute([{:get_upvalue, dest, index} | rest], regs, upvalues, proto, state, cont, frames, line) do
     cell_ref = elem(upvalues, index)
@@ -670,8 +686,6 @@ defmodule Lua.VM.Executor do
     regs = put_elem(regs, dest, value)
     do_execute(rest, regs, upvalues, proto, state, cont, frames, line)
   end
-
-  # ── set_upvalue ────────────────────────────────────────────────────────────
 
   defp do_execute([{:set_upvalue, index, source} | rest], regs, upvalues, proto, state, cont, frames, line) do
     cell_ref = elem(upvalues, index)


davydog187:
## Automated review (workflow `ship-perf-issues`) — stable in round 1

This PR was opened and reviewed by an AI workflow. Posting the reviewer's own findings for the human record.

⚠️ **This is a negative-result / documentation-only PR, not a speedup.** The implementer concluded that #313's premise — inlining depth-1 upvalue *values* into the closure tuple — is infeasible in tight scope, and documented why instead of optimizing.

**Verdict: approve** (as a doc change). The reviewer verified the `get_upvalue`/`set_upvalue` opcode bodies are byte-identical to main (17 insertions, 3 deletions, all comments), so there's no new failure mode, and independently confirmed the rationale is accurate:
- the cell ref is a genuine cross-module contract (`dispatcher.ex`, `_ENV` rebinding in `stdlib.ex`, `debug.getupvalue`/`setupvalue`, `set_open_upvalue`);
- `set_upvalue` can't write back into the immutable closure tuple, so an inlined value would go stale — even read-only inlining is unsound because `debug.setupvalue` can mutate any upvalue at runtime.

Two **nit** findings:
- the merged `# ── get_upvalue / set_upvalue ──` banner drops the standalone `set_upvalue` anchor that the surrounding one-banner-per-opcode layout uses for grep-navigation;
- the comment cites luerl private internals (`get_env_var_1/4` in `luerl_emul.erl`) that may drift across luerl versions — consider softening or pinning the observed version.

**Maintainer decision needed:** close #313 as not-actionable-as-specified, or open a separate design for a closure-value rewrite (per-closure mutable cells) — per the implementer, the only path to a real win.

Issue #324, opened 2026-06-02
Deep recursion ~25% slower since rc.0 — cheapen per-call depth-limit bookkeeping


davydog187:
## Summary

The `string.format` / `table` / dispatcher performance pass landing in `v1.0.0-rc.1` was a net win across the benchmark suite, but the `fibonacci` workload (fib(30), pure recursion) **regressed ~25%** versus `v1.0.0-rc.0`:

* rc.0: 1.64 ips (608.6 ms), ±0.69%
* main/rc.1: 1.23 ips (814.2 ms), ±0.42%

Measured with full Benchee runs on a single machine, run sequentially. Luerl and PUC-Lua (identical code in both runs) were used as drift controls and moved within ±3%, so net of machine drift the regression is still ≈ −22%. This is real, not noise.

## Cause

PR #283 ("configurable max call depth", commit `3a4a0c8`) added per-call bookkeeping to the executor's call/return paths: a `State.check_call_depth!/1` function call plus `call_depth + 1` / `- 1` state updates on **every** Lua function call. `fibonacci` is ~2.7M calls with almost no work per call, so it pays that overhead with nothing to amortize against.

Workloads that do real work per call are unaffected or faster (OOP +41%, closures +4%) because the dispatcher gains (#275, #277) dominate there. The regression is specific to call-dense, work-light code.

## Options to investigate

* Inline the depth check as a plain integer comparison rather than a function call into `State`.
* Check depth every Nth frame instead of on every call.
* Derive depth from the existing `call_stack` length instead of maintaining a separate counter.

## Acceptance criteria

* Recover most of the regression — target within ~5% of rc.0 on `fibonacci` — **without** weakening the call-depth limit.
* Verify with `mix lua.bench --workload fibonacci` before/after on the same machine.

## References

* PR #283 (the cause)
* Benchmark report: `benchmarks/results/2026-06-02-rc0-vs-main.md`
* Called out under "Known issues" in the `v1.0.0-rc.1` CHANGELOG; to be fixed before `1.0.0` final.

davydog187:
Fixed in #359.

Heads-up on the root cause: bisecting `fib(30)` showed the premise here is stale. The #283 depth bookkeeping is **no longer** the bottleneck — removing it entirely from current `main` recovers ~0%, and inlining the check (the suggested fix) measures flat. Later commits absorbed its cost.

[...]

PR #275, opened 2026-05-28
perf(vm): dispatcher coverage for table opcodes and :numeric_for

davydog187:
## Dispatcher table opcodes — make table-heavy workloads bypass the interpreter

Plan: [`.agents/plans/B5b-v2-dispatcher-tables.md`](.agents/plans/B5b-v2-dispatcher-tables.md)
Closes #271

### Goal

Extend `Lua.VM.Dispatcher` and `Lua.Compiler.Bytecode` to lower the
table opcode family plus `:numeric_for`. After this PR all four
`table_ops` benchmarks compile end-to-end and stay out of the
interpreter fallback path, plus the `closures` orchestrator
(`run_closures`).

### Success criteria

- [x] `Lua.Compiler.Bytecode` learns encoders for `:new_table`,
      `:get_table`, `:set_table`, `:set_field`, `:set_list` (basic
      non-multi-return form), `:length`, and `:numeric_for` (body
      encoded recursively; falls back if body contains an uncovered
      opcode or `:break`).
- [x] `Lua.VM.Dispatcher` gains one `case` branch per new opcode
      plus the `:cps_for` continuation marker handled by
      `finish_body/6`.
- [x] `:call` with `result_count == 0` (statement-call form) is
      covered via a new `@op_call_zero` tag — required so
      `run_table_sort` can compile (`table.sort(t)` is a statement
      call).
- [x] `Lua.VM.Executor` exposes six new `dispatcher_*` public
      bridges (`dispatcher_get_table`, `dispatcher_set_table`,
      `dispatcher_set_field`, `dispatcher_length`,
      `dispatcher_coerce_numeric_for_controls`,
      `dispatcher_close_open_upvalues_at_or_above`) — each wraps an
      existing `defp` helper so metamethod fidelity matches the
      interpreter for free.
- [x] `mix compile --warnings-as-errors` passes.
- [x] `mix test`: **1883 → 1902 tests, 0 failures, 25 skipped**.
- [x] `mix test --only lua53`: 29 tests, 0 failures, 18 skipped.
- [x] `test/lua/vm/leak_regression_test.exs`: 3 tests, 0 failures.
- [x] All four `table_ops` benchmark functions compile to bytecode
      end-to-end (verified by a new dispatcher-test setup that asserts
      `{:compiled_closure, _, _}` for each global after install).
- [x] `run_closures` orchestrator compiles (`make_counter` still
      falls back because of `:closure` — that's B5c-v2).
[...]

PR #326, opened 2026-06-02
perf(vm): inline per-call depth check on hot call/return paths

davydog187:
## What

Cheapens the per-call recursion-depth bookkeeping that PR #283
("configurable max call depth", `3a4a0c8`) added to every Lua
call/return. That PR put a `State.check_call_depth!/1` **cross-module
call** plus a `call_depth + 1` / `- 1` counter update on the hottest path
in the VM.

This change replaces the cross-module guard with a module-local
`next_call_depth!/1` marked `@compile {:inline}` in both
`Lua.VM.Dispatcher` and `Lua.VM.Executor`. For the default
`max_call_depth: :infinity` case the check collapses to a plain integer
increment with no call setup. The helper returns the already-incremented
depth, so the call sites no longer read `call_depth` a second time to bump
it. `State.next_call_depth!/1` stays as the authoritative definition the
inlined copies mirror (and the entry point for any non-hot-path caller).

This implements the first option listed in #324 (inline the depth check
as a plain integer comparison rather than a call into `State`).

## How correctness is preserved

The depth **limit** is not weakened:

- `next_call_depth!/1` raises the same catchable Lua `"stack overflow"`
  at exactly the same frame whenever a finite `max_call_depth` is reached
  — the comparison (`depth < max`) and the raise clause are byte-for-byte
  the same logic as the old `check_call_depth!/1`.
- The counter still increments on every push and decrements on every pop
  (call and return), so it unwinds to 0 on normal return and is restored
  after a `pcall`-caught overflow.
- `test/lua/vm/recursion_depth_test.exs` and
  `test/lua/error_gallery_test.exs` (the stack-overflow gallery case)
  pass unchanged, as does the full suite (2153 passed, 19 skipped) and
  the Lua 5.3 official suite embedded in it.

## Benchmark (this machine, drift-controlled)

`fibonacci` = fib(30), ~2.7M near-empty recursive calls. Measured with
Benchee (8s, 2s warmup), alternating rc.1 vs this branch in the same
session with **luerl as a drift control** (luerl held at ~748 ms across
all runs). `lua (chunk)`, 3-round average:

| build              | avg ms | ips  | vs luerl |
|--------------------|--------|------|----------|
| rc.0 (target)      | 642.7  | 1.56 | 0.86x    |
| rc.1 / main (before)| 838.7 | 1.18 | 1.12x    |
| **this branch (after)** | **807.6** | **1.24** | **1.08x** |

Delta: **-31 ms (~3.7%)**, ips 1.18 -> 1.24.

## Why this does not fully reach the rc.0 target

#324 targets "within ~5% of rc.0" (642 ms). I instrumented this directly:
building a throwaway variant that strips **all** depth bookkeeping from
the hot paths (no check, no counter) only reaches ~744 ms on this machine
— still ~100 ms above rc.0. So roughly half of the rc.0 -> rc.1
deep-recursion regression is depth-attributable and half comes from other
rc.1 dispatcher changes (e.g. #275 / #277) that are out of scope for this
issue and benefit the work-heavy workloads.

Within the depth-attributable slice, the cross-module call was the cleanly
removable cost; the residual (counter field writes + the inlined compare)
profiled at the noise floor, and removing it would mean branching the
call/return struct updates on `:infinity` — extra complexity for a
sub-noise gain. I kept the change tight and correct rather than chase it.

See the `flags` / discussion: the remaining gap is tracked by the broader
rc.1 dispatcher work, not weakenable depth limits.

### DIFF
--- a/lib/lua/vm/dispatcher.ex
+++ b/lib/lua/vm/dispatcher.ex
@@ -32,9 +32,15 @@ defmodule Lua.VM.Dispatcher do
 
   # Per-call depth check, inlined to a plain integer comparison so the
   # call-dense recursive path does not pay a cross-module call on every
-  # Lua call/return. Mirrors `State.next_call_depth!/1`: returns the
-  # bumped depth, or raises a catchable "stack overflow" once the limit
-  # is hit. The `:infinity` head (the default) skips the comparison.
+  # Lua call/return. Returns the bumped depth, or raises a catchable
+  # "stack overflow" once the limit is hit. The `:infinity` head (the
+  # default) skips the comparison.
+  #
+  # This is one of three copies of the same logic: the authoritative
+  # `Lua.VM.State.next_call_depth!/1` and an identical private copy in
+  # `Lua.VM.Executor`. They are not compiler-checked against each other,
+  # so any change here must be mirrored in the other two. The shared
+  # semantics are pinned by `test/lua/vm/state_test.exs`.
   @compile {:inline, next_call_depth!: 1}
 
   defp next_call_depth!(%State{call_depth: depth, max_call_depth: :infinity}), do: depth + 1
--- a/lib/lua/vm/dispatcher.ex
+++ b/lib/lua/vm/dispatcher.ex
@@ -30,6 +30,20 @@ defmodule Lua.VM.Dispatcher do
   alias Lua.VM.Table
   alias Lua.VM.Value
 
+  # Per-call depth check, inlined to a plain integer comparison so the
+  # call-dense recursive path does not pay a cross-module call on every
+  # Lua call/return. Mirrors `State.next_call_depth!/1`: returns the
+  # bumped depth, or raises a catchable "stack overflow" once the limit
+  # is hit. The `:infinity` head (the default) skips the comparison.
+  @compile {:inline, next_call_depth!: 1}
+
+  defp next_call_depth!(%State{call_depth: depth, max_call_depth: :infinity}), do: depth + 1
+  defp next_call_depth!(%State{call_depth: depth, max_call_depth: max}) when depth < max, do: depth + 1
+
+  defp next_call_depth!(%State{call_stack: call_stack}) do
+    raise RuntimeError, value: "stack overflow", call_stack: call_stack
+  end
+
   # Mirror of the interpreter's concat ceiling, inlined as a compile-time
   # constant so the binary-binary fast path stays a single comparison. See
   # `Lua.VM.Executor.concat_checked/2` for the rationale.
@@ -582,12 +596,12 @@ defmodule Lua.VM.Dispatcher do
               {code, pc + 1, regs, upvalues, proto, cont, :discard, state.open_upvalues}
 
             call_info = Executor.dispatcher_call_info(proto, name_hint, 0)
-            State.check_call_depth!(state)
+            depth = next_call_depth!(state)
 
             state = %{
               state
               | call_stack: [call_info | state.call_stack],
-                call_depth: state.call_depth + 1,
+                call_depth: depth,
                 open_upvalues: %{}
             }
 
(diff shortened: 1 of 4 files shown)


davydog187:
## Automated review (workflow `ship-perf-issues`) — stable in round 1

This PR was opened and reviewed by an AI workflow. Posting the reviewer's own findings for the human record.

**Verdict: approve.** Behavior-preserving refactor that inlines the per-call depth check on the 8 hot-path push sites in `Dispatcher` and `Executor`, collapsing the default `:infinity` case to a plain integer increment with no cross-module call. The reviewer confirmed depth-limit enforcement is byte-identical to the old `check_call_depth!/1` across `call_zero`/`call_one`/`call_multi`/`tailcall`, decrement/pop sites untouched, full suite green (2153 passed), clean under `--warnings-as-errors`. No correctness regressions.

Two **low** (quality-only) findings:
- **`next_call_depth!/1` is now triplicated** across `State` (public), `Dispatcher` (private), and `Executor` (private). The inlined copies are a deliberate hot-path tradeoff, but a future edit to one copy wouldn't be caught by the compiler — the copies are only validated transitively through end-to-end recursion tests. Suggest a cross-reference note at each copy.
- **`State.next_call_depth!/1` has no in-repo callers and no direct unit test.** A small direct test (bumped value, under-limit value, `:infinity` skip, raise-at-limit with `"stack overflow"`) would pin the authoritative copy and give the three definitions a shared spec.

### Scope note carried from the implementer
Recovers ~3.7% of the fib regression, **not** the issue's "within ~5% of rc.0" target. A throwaway experiment stripping *all* depth logic still measured ~744 ms vs rc.0's ~642 ms — so only ~half the regression is depth-attributable; the rest comes from other rc.1 dispatcher refactors (#275/#277) outside this issue's scope. Worth a maintainer call on whether to keep #324 open for that residual.

davydog187:
Lets address these two

> next_call_depth!/1 is now triplicated across State (public), Dispatcher (private), and Executor (private). The inlined copies are a deliberate hot-path tradeoff, but a future edit to one copy wouldn't be caught by the compiler — the copies are only validated transitively through end-to-end recursion tests. Suggest a cross-reference note at each copy.
State.next_call_depth!/1 has no in-repo callers and no direct unit test. A small direct test (bumped value, under-limit value, :infinity skip, raise-at-limit with "stack overflow") would pin the authoritative copy and give the three definitions a shared spec.

davydog187:
## Automated feedback pass (workflow address-pr-feedback-until-stable) — stable after 1 round(s)

**Final state: STABLE** — both maintainer feedback items addressed; reviewers found no blocking issues.

### Round 1

**Pushed:**
- `429a219` — test(vm): pin next_call_depth! and cross-reference its inlined copies

**Validation:**
- `mix format --check-formatted`: clean (exit 0)
- `mix compile --warnings-as-errors`: lua app clean (only external dependency warnings)
- `mix test`: 2158 passed (61 doctests, 51 properties, 2046 tests), 19 skipped, 1 excluded, 0 failures (+5 new State tests vs 2153 prior)
- Targeted: `test/lua/vm/state_test.exs` 5 passed; `recursion_depth_test.exs` passed unchanged

**Reviewer verdicts:** 2 re-reviews, both STABLE. No high or medium findings. One low/nit:
- _Low_ — New unit test does not directly assert the off-by-one boundary (`call_depth: max-1 -> max`); predicate `depth < max` is correct and the boundary is exercised transitively by `recursion_depth_test`.

### Feedback items

1. **Add cross-reference notes at the three triplicated `next_call_depth!/1` copies (do not un-inline).** Resolved: the authoritative `Lua.VM.State.next_call_depth!/1` `@doc` now states the two private copies in `Lua.VM.Dispatcher` and `Lua.VM.Executor` mirror it, that those copies are not compiler-checked, and that `test/lua/vm/state_test.exs` pins the shared semantics. Each private copy's comment points back at `State` plus its sibling private copy and the test. Inlining and `@compile {:inline}` directives left untouched.

2. **Add a direct unit test for `State.next_call_depth!/1` (bumped value, under-limit, `:infinity` skip, raise-at-limit with "stack overflow"); consider a shared spec.** Resolved: added `test/lua/vm/state_test.exs` (async) with 5 tests covering bumped under-limit (3->4), `:infinity` (41->42), `:infinity` skip at default depth 0 (->1), raise at the limit with `value: "stack overflow"` and the original `call_stack`. The test moduledoc plus the three cross-reference notes serve as the shared spec. No code behavior changed.

PR #327, opened 2026-06-02
perf(vm): honest max_registers — drop register slack on both VM engines

### LINKED ISSUE #312: perf(vm): drop +16 register buffer on interpreter :call path
## Context

Fibonacci (recursive) is **1.18× slower than luerl** (1.23s vs 1.04s in
`benchmarks/fibonacci.exs`).

Internal signal: `benchmarks/dispatcher_vs_interpreter.exs` shows the
dispatcher path running fib(25) **1.20× faster** than the interpreter. The
dispatcher allocates exactly `max(max_registers, param_count)` per frame
(`lib/lua/vm/dispatcher.ex:1375-1381`); the interpreter allocates **+16
extra slots** per frame (`lib/lua/vm/executor.ex:978-979`):

\`\`\`elixir
callee_regs =
  Tuple.duplicate(nil, max(callee_proto.max_registers, param_count) + 16)
\`\`\`

For fib(28) (~1.4M frames), those 16 extra slots per frame are pure
allocation churn.

## Change

`lib/lua/vm/executor.ex:978-979`.

Audit the call sites that historically required the buffer (multi-return
inline expansion into the callee's register window). Then either:

a) Prove codegen reports honest peak register usage (matching the
   dispatcher's codegen contract) and **remove the buffer entirely**.
b) Keep a much smaller buffer (2–4) gated on `is_vararg` / known
   multi-return callees only.

The dispatcher already runs with no buffer, which establishes the safety
argument for at least the dispatcher-eligible subset.

## Verification

- **Full** `mix test` — miss-sized callee tuples manifest as runtime errors,
  not static failures.
- Lua 5.3 official suite runner (whatever script is canonical).
- `mix run benchmarks/fibonacci.exs`
- `mix run benchmarks/dispatcher_vs_interpreter.exs` — gap should close,
  not invert.

## Expected

Noticeable share of the 1.18× fib gap and the closures-benchmark gap.

davydog187:
## Summary

Makes codegen's `max_registers` an **honest** register peak and sizes both VM engines' register tuples to it with **no slack buffer**. This closes #312 (drop the interpreter's `+16` buffer) and unifies that with the dispatcher-side fix merged in #359 — so there is now **one register-sizing source of truth** across the interpreter and the dispatcher.

> Rebased onto current `main` (post-#359) and reconciled. See "Relationship to #359" below.

## Background

Both engines over-allocated every call frame's register tuple to mask the same latent bug:
- The **interpreter** added a `+16` buffer in `call_function/3`, the `:call` opcode, and `call_value/5`.
- The **dispatcher** added a `+16` slack in `init_callee_regs/4` (#347), which regressed call-dense workloads ~25%+ (#324).

The buffers existed because `peak_reg`/`next_reg` tracking in codegen could **undercount** `max_registers` for a fixed destination register in deeply nested constructs (e.g. a concat chain inside a table-constructor list field nested in several `for` bodies emitted `{:concatenate, 53, _, _}` while reporting `max_registers = 53`). These shapes contain multi-return calls inside `set_list`, which the bytecode encoder rejects, so they only ever execute on the interpreter — which is why the dispatcher could already run buffer-free and the interpreter could not.

## The fix

Make `max_registers` honest at the root, then drop both buffers:

- **`instruction_peak/1`** (new, in `codegen.ex`) walks the emitted instruction stream — recursing into `numeric_for` / `generic_for` / `while` / `repeat` / `test` / short-circuit bodies — and returns the true high-water register index + 1. It is folded into `max_registers` at both prototype-finalization sites.
- Both engines now size to **`max(max_registers, param_count)`** with no buffer. Dynamic-range writers (`:call` results, unbounded `:vararg`, multi-return distribution) still grow the tuple lazily — `ensure_regs_capacity/2` in the interpreter, `grow_regs/2` in the dispatcher. The interpreter's unbounded `:vararg` (`count == 0`) site now grows before writing, matching the dispatcher.

## Relationship to #359 (merged)

#359 recovered the #324 dispatcher regression by having the dispatcher compute its **own** exact `reg_file_size` by walking the emitted bytecode — a workaround for the codegen undercount. Now that `instruction_peak/1` makes `max_registers` honest, that separate walker is redundant. This PR **removes it** (drops the `reg_file_size` field and the `max_register_used`/`register_positions` walker from `Lua.Compiler.Bytecode`) and points the dispatcher back at `max(max_registers, param_count)`. The `max_registers` invariant test keeps its independent bytecode-level walker as a cross-check against `instruction_peak/1` and now asserts against `max_registers`.

Net: the dispatcher's runtime path is byte-identical to #359 (same tuple sizes), and the interpreter joins it.

## Correctness

- `mix test` (default + `--include slow --include differential --include lua53`, the official Lua 5.3 suite): **2374 passed**, 19 skipped.
- The generative short-circuit harness that overflowed when the buffer was first removed now passes (its register peak is covered by `instruction_peak/1`).
- `mix format --check-formatted` clean; compiles with `--warnings-as-errors`.

## Benchmarks (Apple M-series, drift-controlled by Luerl)

- `dispatcher_vs_interpreter` fib(25): **interpreter ~95 ms → ~72 ms** (~26% faster); dispatcher–interpreter gap closes to ~1.02× (not inverted).
- `fibonacci` fib(30), dispatcher path: ~1.125× slower than Luerl — unchanged from #359 (the dispatcher path is byte-identical), confirming no regression from the reconciliation.
- No other workload slower.

Closes #312.


davydog187:
## Automated review (workflow `ship-perf-issues`) — 2 rounds, now stable

This PR was opened and reviewed by an AI workflow. Posting the reviewers' own findings for the human record.

### Round 1 — HIGH (fixed in `ef60cea`)
**`instruction_size` backstop omitted `:greater_than` / `:greater_equal`** (`lib/lua/compiler/codegen.ex`)

The backstop that replaces the dropped +16 buffer must count *every* statically-fixed destination register. Codegen emits `{:greater_than, dest, ..}` / `{:greater_equal, dest, ..}` (siblings of `:lt`/`:le`, which *are* in the allowlist), but both fell through to the `do: 0` catch-all. A `>`/`>=` whose destination is the function's high-water register — inside the same deeply-nested shape that motivated this PR — would under-count `max_registers` by one, with no buffer left to absorb it: a runtime `ArgumentError` on `put_elem`, not a compile failure. Exactly the #312 failure class.
→ **Addressed** by the follow-up commit `perf(vm): count > and >= destinations in max_registers backstop`.

Round 1 also flagged (low): the `instruction_size(_instr), do: 0` catch-all silently hides future register-writing opcodes — the same fragility that produced this bug.

### Round 2 — stable, quality-only findings remain
- **low — `lib/lua/vm/vm.ex:22`**: the top-level chunk still allocates `max_registers + 16` with a comment that now contradicts the PR's thesis. Not a regression (extra slack is safe), but it means the new `instruction_peak` backstop is *not* exercised on the main-chunk path — every adversarial probe had to be wrapped in `local function f() … end` to hit the buffer-free path. Either drop the +16 here too, or rewrite the comment to say it's intentional defense-in-depth.
- **low**: two stale dispatcher comments + dead `:scope` handling (cleanup).

### Verification reported by the reviewers
Full suite 2153 passed / 19 skipped; the `:slow` Lua 5.3 official suite passed; a 400–600 case fuzz of nested concat/arith in table constructors inside nested `for` loops on the buffer-free `:call` path produced zero register-overflow crashes. Benchmark: interpreter fib(25) ~21% faster; dispatcher/interpreter gap closed from 1.21× → 1.03× (not inverted).

lib/lua/compiler/codegen.ex
```diff
@@ -87,6 +93,107 @@ defmodule Lua.Compiler.Codegen do
...
+                :less_equal,
+                :greater_than,
+                :greater_equal,
+                :not_equal,
+                :not,
+                :length,
+                :closure
+              ] do
+    elem(instr, 1) + 1
+  end
```
davydog187: there's got to be a better way to do this. we're duplicating all of these instructions with nothing tying it back to changes over time.

davydog187: Addressed in 5036f34. The duplication can't fully go away — the interpreter runs on the pre-encoding instruction stream, so this walker has to live alongside the bytecode-level one in the invariant test — but the *silent* failure mode is gone. The `do: 0` catch-all now raises: every store/control/debug op is classified explicitly as zero, and any unclassified instruction fails loudly at compile time (the first program to emit it). That's what ties it to changes over time — you can no longer add a register-writing opcode and have it quietly undercount `max_registers` (the way `>`/`>=` did). A new guard test (`instruction_size_test.exs`) drives a sample of every emitted opcode through the walker and pins its slot count.

davydog187: Correction — reworked in 2d1eb3f (force-pushed over the earlier commit). The earlier pass just wrapped the allowlist in a raising catch-all; the allowlist itself was the problem, so it's gone now. The walker enumerates only the genuine exceptions (range/offset writers, nested-body recursers, and value-less ops — stores/returns/breaks/gotos/labels/source lines) and **defaults everything else to "destination is operand 1"** — the rule the 34-item list was spelling out one opcode at a time. A new load/arith/comparison opcode is now sized correctly with zero changes here. And the direction is safe: the default can only ever over-count by a slot (benign slack), never undercount into the #312 crash, since the only thing that could go wrong is forgetting a value-less op.

PR #333, opened 2026-06-04
fix(vm): keep heap effects across protected-call error unwinding

### LINKED ISSUE #331: pcall discards state mutations made before error() (diverges from reference Lua)
## Summary

When a function called via `pcall` raises an error, any state mutations made before the `error()` call are rolled back. Reference Lua keeps those mutations; `pcall` only traps the error and returns `false, err` without unwinding side effects.

## Reproduction

```elixir
Mix.install([{:lua, "~> 1.0.0-rc.1"}])

code = """
x = 1
local ok, err = pcall(function()
  x = 2
  error("boom")
end)
return x, ok, err
"""

{results, _lua} = Lua.eval!(Lua.new(), code)
IO.inspect(results)
```

**Actual:** `[1, false, "boom"]`
**Expected:** `[2, false, "boom"]`

Reference Lua 5.x for comparison:
```
$ lua -e 'x = 1; pcall(function() x = 2; error("boom") end); print(x)'
2
```

## Why this matters

Standard Lua code routinely relies on this:
- Counters incremented before a failed call (`step = step + 1; risky()`)
- Partial progress recorded for retry/resume logic
- "Mutate then validate, error on bad result" patterns
- Logging or audit trail writes inside the protected block

All of these silently lose their effects.

## Root cause

In `lib/lua.ex`, `do_call_function/3`'s rescue clause returns the original `state` parameter rather than the partially-mutated state accumulated during execution:

```elixir
defp do_call_function({:lua_closure, proto, upvalues}, args, state) do
  ...
  {results, _regs, new_state} =
    Executor.execute(proto.instructions, callee_regs, upvalues, proto, state)

  {:ok, results, new_state}
rescue
  e -> {:error, Exception.message(e), state}   # <- should thread out the partial new_state
end
```

The same pattern repeats in the `:native_func` and `:compiled_closure` clauses. To match reference Lua, the executor needs to thread the mutated state out through the raise (e.g., attach it to a struct exception) so the rescue can return `{:error, msg, partial_state}` instead of the pre-call snapshot.

## Reproducible on

`lua` v1.0.0-rc.1 (also v1.0.0-rc.0).

davydog187:
Closes #331.

## Problem

When a function called via `pcall`/`xpcall` (or `Lua.call_function/3` from Elixir) raised an error, every mutation made before the error — global writes, table field updates, upvalue assignments, metatable changes — was rolled back. Reference Lua keeps those heap effects and only unwinds control state (Lua 5.3 §2.3):

```
$ lua -e 'x = 1; pcall(function() x = 2; error("boom") end); print(x)'
2
```

Root cause: all VM state threads through the immutable `%Lua.VM.State{}` struct, and errors are Elixir exceptions that carried no state — so a protected call's `rescue` only had its stale entry snapshot.

## Fix

VM exceptions (`RuntimeError`, `TypeError`, `AssertionError`, `ArgumentError`) gain a `:state` field carrying the raise-time state. Protected boundaries recover it via the new `State.unwind_to/2`: **heap fields** (tables, userdata, metatables, upvalue cells, private) come from the raise-time snapshot; **control fields** (call stack, call depth, open upvalues) restore from the entry state.

State is attached in layers so no raise path is missed, with zero happy-path overhead:

1. **Raise sites with state in scope** — `error()`, `assert()`, stack overflow, call-type errors (both engines) annotate directly.
2. **Native-call choke point** — both native invocation sites reraise escaping VM exceptions annotated with the call's entry state, covering every stdlib bad-argument check without touching ~200 raise sites. Innermost annotation wins.
3. **In-frame pure helpers** — arithmetic/comparison/concat/index/bitwise/length/numeric-for helpers now take the caller's threaded state (one extra already-bound argument; raise opts only built on the error path), so `x = 2; return nil + 1` under pcall keeps `x = 2`.
4. **Frame-boundary backstop** — frame entries in both engines annotate-if-nil, bounding any future missed raise site to losing only the innermost frame's in-frame mutations. One `try` per frame entry; the executor loop's tail recursion is untouched.

The `xpcall` message handler now runs against the recovered state, so it observes the mutations (matching PUC-Lua). `Lua.call_function/3`'s `{:error, reason, lua}` likewise returns the recovered state. VM exceptions derive `Inspect` excluding `:state` so logging a trapped error never dumps the VM heap.

## TDD

The first commit is a red 37-case matrix (`test/lua/vm/pcall_state_preservation_test.exs`) pinning reference semantics — global/table/upvalue/metatable mutations crossed with `error()`, `assert`, runtime type errors, stdlib bad-argument errors, stack overflow, nested pcall, xpcall (including an erroring handler), and the Elixir API — each run under **both engines** (dispatcher and bytecode-stripped interpreter). Expected values were cross-checked against PUC Lua. Each subsequent commit turns a slice green; control tests pin that no-mutation and success paths are unchanged.

## Verification

- `mix test`: 2190 passed, 0 failures (includes the official Lua 5.3 suite tests)
- Issue repro now returns `[2, false, "boom"]` (was `[1, false, "boom"]`)
- Focused benchmark vs `main` (benchmarks/scripts/pcall_state_bench.exs): fib(26) −0.2%, table-write 200k −6.6%, concat 5k +1.2%, pcall loop 50k +0.4% — all within noise

## Notes

- `pcall`'s rescue list now explicitly includes `ArgumentError` (previously trapped by the catch-all clause; message behavior unchanged).
- Out of scope, observed while testing: `error({code = 1})` stringifies the error object instead of passing it through to pcall's second return — now tracked in #334.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- **Protected calls no longer roll back heap effects** (#331). When a function
+  called via `pcall`/`xpcall` (or `Lua.call_function/3` from Elixir) raised an
+  error, mutations made before the error — global writes, table field updates,
+  upvalue assignments, metatable changes — were silently discarded, diverging
+  from reference Lua. VM exceptions now carry the raise-time state, and
+  protected-call boundaries recover it: heap state is kept, control state
+  (call stack, open upvalues) unwinds. The `xpcall` message handler now also
+  observes those mutations, matching PUC-Lua's handler semantics.
+
 ## [v1.0.0-rc.1] - 2026-06-02
 
 The second release candidate for `1.0.0`. It builds on rc.0 with a new
--- a/benchmarks/scripts/pcall_state_bench.exs
+++ b/benchmarks/scripts/pcall_state_bench.exs
@@ -7,6 +7,12 @@
 #   * table writes   — t[i] = i loop (table_newindex fast path)
 #   * concat loop    — concat_checked/concat_coerce widening
 #   * pcall loop     — protected-call overhead incl. the new try wrappers
+#   * closure for    — generic_for driven by a Lua-closure iterator, which is
+#                      the path that gained a new per-iteration try block in
+#                      executor.ex call_value/5 (lua_closure clause)
+#
+# memory_time is enabled so the per-call allocation/GC delta of the added try
+# frames is observable, not just wall-clock time.
 
 defmodule PcallStateBench do
   @moduledoc false
@@ -51,14 +57,32 @@ pcall_loop =
   return n
   """)
 
+# A stateful Lua-closure iterator drives the generic_for loop, so each
+# iteration routes through executor.ex call_value/5's lua_closure clause —
+# the path that gained a new per-iteration try block in this change.
+closure_for =
+  PcallStateBench.chunk(lua, """
+  local function range(limit)
+    local i = 0
+    return function()
+      i = i + 1
+      if i <= limit then return i end
+    end
+  end
+  local sum = 0
+  for v in range(50000) do sum = sum + v end
+  return sum
+  """)
+
 Benchee.run(
   %{
     "fib(28)" => fn -> Lua.eval!(lua, fib) end,
     "table writes 200k" => fn -> Lua.eval!(lua, table_writes) end,
     "concat 5k" => fn -> Lua.eval!(lua, concat_loop) end,
-    "pcall loop 50k" => fn -> Lua.eval!(lua, pcall_loop) end
+    "pcall loop 50k" => fn -> Lua.eval!(lua, pcall_loop) end,
+    "closure for 50k" => fn -> Lua.eval!(lua, closure_for) end
   },
   warmup: 1,
   time: 4,
-  memory_time: 0
+  memory_time: 2
 )
--- a/benchmarks/scripts/pcall_state_bench.exs
+++ b/benchmarks/scripts/pcall_state_bench.exs
@@ -1,0 +1,63 @@
+# Focused micro-benchmark for the pcall heap-effect preservation change.
+# Run on two checkouts and diff:
+#   MIX_ENV=benchmark mix run benchmarks/scripts/pcall_state_bench.exs
+#
+# Covers the paths the change touches:
+#   * fib(28)        — arithmetic hot path (widened safe_* helper arities)
+#   * table writes   — t[i] = i loop (table_newindex fast path)
+#   * concat loop    — concat_checked/concat_coerce widening
+#   * pcall loop     — protected-call overhead incl. the new try wrappers
+
+defmodule PcallStateBench do
+  def chunk(lua, code) do
+    {chunk, _} = Lua.load_chunk!(lua, code)
+    chunk
+  end
+end
+
+lua = Lua.new()
+
+{_, lua} =
+  Lua.eval!(lua, """
+  function fib(n)
+    if n < 2 then return n end
+    return fib(n-1) + fib(n-2)
+  end
+  """)
+
+fib = PcallStateBench.chunk(lua, "return fib(28)")
+
+table_writes =
+  PcallStateBench.chunk(lua, """
+  local t = {}
+  for i = 1, 200000 do t[i] = i end
+  return #t
+  """)
+
+concat_loop =
+  PcallStateBench.chunk(lua, """
+  local s = ""
+  for i = 1, 5000 do s = s .. "x" end
+  return #s
+  """)
+
+pcall_loop =
+  PcallStateBench.chunk(lua, """
+  local n = 0
+  for i = 1, 50000 do
+    local ok = pcall(function() n = n + 1 end)
+  end
+  return n
+  """)
+
+Benchee.run(
+  %{
+    "fib(28)" => fn -> Lua.eval!(lua, fib) end,
+    "table writes 200k" => fn -> Lua.eval!(lua, table_writes) end,
+    "concat 5k" => fn -> Lua.eval!(lua, concat_loop) end,
+    "pcall loop 50k" => fn -> Lua.eval!(lua, pcall_loop) end
+  },
+  warmup: 1,
+  time: 4,
+  memory_time: 0
+)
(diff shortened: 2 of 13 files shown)


davydog187:
## Automated multi-agent review

This is an automated multi-agent review of PR #333 (`fix(vm): keep heap effects across protected-call error unwinding`, closes #331). A review-fix loop ran 4 rounds; each finding below was independently confirmed by multiple voting agents and then fixed on this branch. The loop hit its max round count without reaching full stability, so a residual class of edge cases may remain (see status note at the end).

Scope note: error-value passthrough / stringification (issue #334) was explicitly out of scope and not reviewed.

### Confirmed findings (all fixed on this branch)

**[HIGH] Native generic-for iterator raise drops current-frame heap mutations**
`lib/lua/vm/executor.ex:2140`
The `:native_func` clause of `call_value/5` (the choke point for generic-for iterators) wrapped the native call in `try/after` for position restore but had no state-annotating `rescue`. A VM exception raised by a native iterator (e.g. `utf8.codes` on invalid UTF-8, or `next`/`pairs`) escaped with `:state == nil` and was only caught by an outer frame backstop that annotated it with the frame's *entry* state, rolling back all heap mutations made before/inside the loop. Asymmetric with the `:call` opcode native path which annotated correctly.
Fixed in `0056ef1` — *fix(vm): annotate native iterator raises with call-time state* (added `reraise annotate_frame_state(e, state)` mirroring the `:call` path).

**[HIGH] deflua/set! user-function error tuple drops heap mutations under pcall**
`lib/lua.ex:1086` (and `lib/lua.ex:277`)
When a `deflua`-style module function (or `set!/3` user function) returned `{:error, reason, %Lua{state: mutated}}`, the wrapper raised `RuntimeError, value: reason` without attaching `returned_lua.state`. The only annotation that fired was the native-call choke point using the pre-call entry state, so under `pcall`/`xpcall` the user function's heap mutations were rolled back. Same class as the gsub finding (callback mutates state, then we raise without ferrying post-callback state).
Fixed in `48c98f7` — *fix(lua): ferry state out of user-function error tuples under pcall* (raises now carry `state: returned_lua.state`).

**[MEDIUM] gsub function-replacement raise drops callback heap mutations**
`lib/lua/vm/stdlib/pattern.ex:237`
`string.gsub` threads state through a function replacement, but when the callback returned an invalid value (e.g. a table), `apply_replacement/4` raised `RuntimeError` with no `:state`, discarding the post-callback state in scope. The native-call boundary then refilled nil with the pre-callback entry snapshot, rolling back the callback's heap effects. Repro: `x=1; pcall(function() return string.gsub('ab','.',function(c) x=2; return {} end) end); return x` yielded `x==1` (reference Lua keeps `x==2`).
Fixed in `ead6148` — *fix(vm): thread state into gsub invalid-replacement raise*.

**[MEDIUM] No coverage for error in a native-stdlib callback under pcall**
`test/lua/vm/pcall_state_preservation_test.exs:173`
No test covered a native stdlib function that calls back into Lua where the Lua callback raises (e.g. `table.sort` with an erroring comparator, or an erroring `pairs`/`ipairs` loop body). This crosses the native-call choke point twice and exercises the "innermost annotation wins" reraise logic that no other case did. A coverage gap, not a live bug — but it left the native reraise annotation unguarded against regression.
Fixed in `debb1ac` — *test(vm): cover native-callback errors under pcall*.

**[LOW] lua_next discards order_tail flush mutation on invalid-key raise**
`lib/lua/vm/stdlib.ex:335`
`lua_next/2` flushes a table's deferred-append `order_tail` into a fresh `state` before checking the key, but the `:invalid_key` branch raised `ArgumentError` without that flushed state or any `:state` annotation. The flush itself is idempotent cache normalization (harmless), but combined with the missing `call_value` rescue it was one more native iterator raise escaping with nil state.
Addressed in `0056ef1` (the native-iterator rescue fix covers the propagation path).

**[LOW] Doc comments reference private/nonexistent annotate_state/2**
`lib/lua/vm/dispatcher.ex:153` (and `lib/lua/vm/executor.ex:184`)
Comments pointed at `Lua.VM.Executor.annotate_state/2`, which was a private alias; the public function actually called was `annotate_frame_state/2`.
Fixed in `ead6148` — comments now reference `annotate_frame_state/2` and the redundant private `annotate_state/2` alias was collapsed across both engines.

**[LOW] Benchmark script fails repo Styler format check**
`benchmarks/scripts/pcall_state_bench.exs:11`
`defmodule PcallStateBench` lacked `@moduledoc false`, which Styler flags when the file is formatted explicitly (benchmarks/ is outside the default formatter scope, so CI was unaffected).
Fixed in `98fa087` — *chore(benchmarks): add @moduledoc false to pcall_state_bench*.

**[LOW] pcall_state_bench omits memory measurement and the generic-for path**
`benchmarks/scripts/pcall_state_bench.exs:63`
`memory_time: 0` measured only wall-clock, missing per-call allocation/GC pressure (the primary observable cost of added `try` frames), and no case routed through `call_value/5`'s lua-closure clause — the one path that gained a genuinely new per-iteration `try`.
Fixed in `42cab9d` — *bench(vm): measure memory and exercise lua-closure generic_for in pcall bench*.

### Refuted / not actioned findings

<details>
<summary>Notable findings raised during the loop but refuted (for transparency)</summary>

- **[HIGH] No coverage for error raised inside a metamethod under pcall** (`pcall_state_preservation_test.exs`) — refuted; existing in-frame helper threading already covers metamethod raise paths.
- **[LOW] Duplicated raise-time state extraction across lua.ex and stdlib.ex** (`lib/lua.ex`) — refuted as not a defect; intentional per-boundary extraction.
- **[LOW] No coverage for a successful pcall following a failed one** (`pcall_state_preservation_test.exs`) — refuted; recovered-state continuation is exercised by existing cases.
- **[LOW] call_value/5 Lua-iterator path gains a per-iteration try block** (`lib/lua/vm/executor.ex`) — refuted as a concern; the try frame is required for correctness and its cost is addressed by the benchmark update above.

Some lower-confidence findings drew a split vote (e.g. the `lua_next` and benchmark items had one dissenting reviewer who considered them harmless/out-of-consequence); they were fixed anyway since the changes are cheap and strictly correct.
</details>

### Status

- `mix format --check-formatted`: **clean** (exit 0)
- `mix test`: **2202 passed** (61 doctests, 51 properties, 2090 tests), 19 skipped, 1 excluded — **0 failures**
- Branch is pushed and in sync with `origin/fix/pcall-state-preservation`.

Note: the review-fix loop reached its maximum round count without converging to full stability, so a residual edge case may remain unsurfaced. All findings confirmed through round 4 have been fixed and the full suite is green.

PR #335, opened 2026-06-04
fix(vm): pass raw error values through pcall and add §6.1 position prefix

### LINKED ISSUE #334: pcall returns stringified error objects instead of passing the error value through
## Summary

Lua 5.3 §6.1: `error(message)` raises an arbitrary Lua value, and `pcall` returns that value as-is as its second result. This implementation stringifies non-string error objects, and omits the position prefix that reference Lua adds to string errors.

Found while testing #331 (PR #333) — split out as a separate issue since it's about the error *value*, not state preservation.

## Reproduction

```elixir
{r, _} = Lua.eval!(Lua.new(), "local ok, err = pcall(function() error({code = 1}) end) return ok, type(err), err")
# actual:   [false, "string", "table: 0x0000000000000C"]
# expected: [false, "table", <the table; err.code == 1>]

{r, _} = Lua.eval!(Lua.new(), "local ok, err = pcall(function() error(42) end) return ok, type(err), err")
# actual:   [false, "string", "42"]
# expected: [false, "number", 42]

{r, _} = Lua.eval!(Lua.new(), "local ok, err = pcall(function() error(\"boom\") end) return ok, type(err), err")
# actual:   [false, "string", "boom"]
# expected: [false, "string", "file:1: boom"]   -- position prefix
```

Reference Lua:

```
$ lua -e 'local ok, err = pcall(function() error({code = 1}) end); print(ok, type(err), err.code)'
false	table	1
$ lua -e 'local ok, err = pcall(function() error(42) end); print(ok, type(err), err)'
false	number	42
$ lua -e 'local ok, err = pcall(function() error("boom") end); print(ok, type(err), err)'
false	string	(command line):1: boom
```

## Why this matters

Structured error objects (`error({code = ..., msg = ...})`) are the standard Lua idiom for machine-readable errors; callers pattern-match on `err.code` after `pcall`. Stringification silently breaks that. The missing position prefix also matters: per §6.1, `error` adds `source:line:` to *string* messages (unless `level` is 0), and lots of code/tests match on that prefix.

## Pointers

- `Lua.VM.RuntimeError` already carries the raw error value in its `:value` field — `lua_error` (`lib/lua/vm/stdlib.ex`) raises it verbatim.
- The stringification happens in `extract_error_message/1` in `lib/lua/vm/stdlib.ex` (pcall/xpcall rescue path), which calls `Value.to_string/1` on non-binary values.
- `error`'s `level` argument (position prefix control, §6.1) appears unimplemented; the prefix behavior belongs to `lua_error`, not to pcall.
- Since #331 (PR #333), the raise-time state survives into pcall's rescue, so returning a `tref` error object is now safe — the referenced table's heap entry is preserved.
- The `xpcall` message handler should receive the same untouched value.

Note `Lua.VM.RuntimeError`'s host-facing *message* formatting ("(error object is a TYPE value)" for non-string/number values) matches PUC-Lua and should stay — this issue is only about the value `pcall`/`xpcall` hand back to Lua code.

davydog187:
## pcall/xpcall pass the raw Lua error value through; error() adds the §6.1 position prefix

Plan: [`.agents/plans/A48-pcall-error-value-passthrough.md`](../blob/fix/pcall-error-value-passthrough/.agents/plans/A48-pcall-error-value-passthrough.md)
Closes #334

### Goal

Per Lua 5.3 §6.1, `error(value)` raises an arbitrary Lua value and `pcall` returns it **as-is** as its second result; `xpcall` hands it to the message handler untouched. Previously `extract_error_message/1` stringified every non-string value (`error({code = 1})` → `"table: 0x..."`). Additionally, string messages now gain the reference `source:line:` position prefix (`error("boom")` → `"file:1: boom"`), suppressed by `level == 0` — all **without disturbing host-facing error rendering**.

### How

- `lua_error` captures the `level` argument and builds the prefixed view into a new **Lua-facing-only** `:lua_value` field on `RuntimeError`; `:value` stays raw, so `message`/`to_map`/`stringify` (and the `(error object is a TYPE value)` host formatting) are provably untouched — no doubled location.
- `extract_error_message/1` → `error_value/1`: `lua_value` → raw `value` (matched on **key presence**, so `error(nil)`/`error(false)` pass through) → `Exception.message` fallback for `ArgumentError`/plain Elixir exceptions.
- Dispatcher parity: native calls inside compiled closures publish a `{nil, source}` position, so the prefix is **suppressed** rather than attributed to a stale outer line (the dispatcher strips per-instruction lines at encode time). The interpreter attributes correctly.

### Success criteria

- [x] New dual-engine matrix `test/lua/vm/pcall_error_value_test.exs` (26 cases, both engines): table/number/boolean/`false`/`nil` passthrough, prefix shape + source-first ordering, level-0 suppression, xpcall raw handler value, erroring-handler fallback to the original raw value, TypeError/bad-argument regression guards — all green (red-first, commit 424c65a)
- [x] `mix test`: 2234 passed, 0 failures; host-message canaries unchanged (`error_gallery_test.exs`, `integration_test.exs:1382`, `lua_test.exs:940`, `runtime_exception_test.exs:34-35`)
- [x] `mix test --only lua53`: 17 passed / 12 skipped — identical to main baseline
- [x] `mix format` + `mix compile --warnings-as-errors` clean (`mix credo` is not a task in this repo — see Discoveries)

### Changes

```
CHANGELOG.md                                   | 13 ++++
lib/lua/vm/executor.ex                         | 17 ++++-
lib/lua/vm/runtime_error.ex                    | 11 +++-
lib/lua/vm/stdlib.ex                           | 69 ++++++++++++++++-----
test/lua/vm/pcall_error_value_test.exs         | 285 +++++++++++++++++++++
test/lua/vm/pcall_state_preservation_test.exs  | 24 +++++--
```

### Discoveries

- The position-suppression edit means native raises (`assert`, bad-argument) inside nested compiled closures now render a source-only host header (`at <eval>:`) where they previously had **no** location header at all (`{nil, nil}` position). Verified via before/after canary: strictly added information, no wrong line, no test pinned the old bytes; `ErrorFormatter.format_location(source, nil)` is an explicit supported clause.
- The plan's verification listed `mix credo`, but credo is not a dependency of this repo.
- Follow-up needed (issue not yet filed): a dispatcher native-call **line bridge** so `error()` inside compiled closures gets a real line (collapsing the engine-conditional prefix assertions) and `error(msg, 2)` caller attribution becomes implementable.

- `error()` level >= 2 (needs per-frame line numbers; treated as level 1)
- Position prefix on internal raises (`TypeError`/`ArgumentError` — host `ErrorFormatter` already renders location; many tests pin exact strings)
- Correct compiled-closure prefix **line** under the dispatcher (suppressed instead; see Discoveries)
- Unskipping `test/lua53_tests/errors.lua` (blocked on an unrelated `load()` parse-error-prefix mismatch; its level-0 and `error()`-nil cases are reproduced in the new unit test)
- Any change to host-facing rendering

### Verification

```
mix format                          # clean
mix compile --warnings-as-errors    # clean
mix test test/lua/vm/pcall_error_value_test.exs   # 26 passed
mix test                            # 2234 passed, 0 failures, 19 skipped
mix test --only lua53               # 17 passed, 12 skipped (== main baseline)
```

### DIFF
--- a/lib/lua/vm/executor.ex
+++ b/lib/lua/vm/executor.ex
@@ -487,8 +487,22 @@ defmodule Lua.VM.Executor do
   def dispatcher_call_function({:compiled_closure, _, _} = closure, args, state, _proto, _name_hint),
     do: call_function(closure, args, state)
 
-  def dispatcher_call_function({:native_func, _} = nf, args, state, _proto, _name_hint),
-    do: call_function(nf, args, state)
+  def dispatcher_call_function({:native_func, _} = nf, args, state, proto, _name_hint) do
+    # The dispatcher strips per-instruction line info at encode time, so any
+    # position visible here is a stale snapshot from an outer executor frame.
+    # Publish `{nil, source}` for the duration of the native call: raise
+    # sites that read `current_position/0` (e.g. `error()`'s §6.1 position
+    # prefix) then omit the line rather than attribute a wrong one. Restored
+    # after the call so nested invocations don't leak.
+    prev_pos = Process.get(@position_key, @unset)
+    set_position(nil, proto.source)
+
+    try do
+      call_function(nf, args, state)
+    after
+      restore_position(prev_pos)
+    end
+  end
 
   def dispatcher_call_function(other, args, state, proto, name_hint) do
     case get_metatable(other, state) do
--- a/.agents/plans/A48-pcall-error-value-passthrough.md
+++ b/.agents/plans/A48-pcall-error-value-passthrough.md
@@ -2,10 +2,10 @@
 id: A48
 title: pcall/xpcall pass the raw Lua error value through; error() adds the §6.1 position prefix
 issue: 334
-pr: null
+pr: 335
 branch: fix/pcall-error-value-passthrough
 base: main
-status: in-progress
+status: review
 direction: A
 unlocks:
   - structured error objects (error({code = ...})) survive pcall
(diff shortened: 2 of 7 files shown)


lib/lua/vm/executor.ex
```diff
@@ -487,8 +487,22 @@ defmodule Lua.VM.Executor do
...
-  def dispatcher_call_function({:native_func, _} = nf, args, state, _proto, _name_hint),
-    do: call_function(nf, args, state)
+  def dispatcher_call_function({:native_func, _} = nf, args, state, proto, _name_hint) do
+    # The dispatcher strips per-instruction line info at encode time, so any
+    # position visible here is a stale snapshot from an outer executor frame.
+    # Publish `{nil, source}` for the duration of the native call: raise
+    # sites that read `current_position/0` (e.g. `error()`'s §6.1 position
+    # prefix) then omit the line rather than attribute a wrong one. Restored
+    # after the call so nested invocations don't leak.
+    prev_pos = Process.get(@position_key, @unset)
```
davydog187, line 497: Process dictionary is really a hack. I think this was from a previous design choice made for performance that we need to revisit

davydog187: we will evaluate this later

PR #340, opened 2026-06-09
feat(os): seed os.clock origin at install, add os.time_ms/os.time_us

### LINKED ISSUE #330: Address issues with os.time
Currently there's two issues with `os`

1. `os.clock()` says it returns time since "boot", but it really is since the first time since its called (we dont actually call it in the install as it claims). I'm not sure what the Lua stdlib expects here, but we should call `boot_offset()` in install to properly seed the value

2. `os.time()` returns seconds, which is probably standard behavior. I think we should either change this to `millisecond` precision, or expose two new functions `os.time_ms()` and `os.time_us()`

davydog187:
Closes #330.

## Context

Issue #330 flagged two problems with the `os` standard library:

1. **`os.clock()` origin was seeded lazily.** The `boot_offset/0` helper's comment claimed the monotonic origin was "captured at module load," but it was actually seeded on the *first call* to `os.clock()` — so the origin drifted to whenever a program first happened to call it.
2. **No sub-second epoch precision.** `os.time()` returns whole seconds (correct PUC-Lua behavior), leaving programs that need finer resolution with no option.

## Changes

- **Seed `boot_offset()` in `install/1`** so `os.clock()` measures elapsed time from a stable startup point. Updated the stale `boot_offset/0` comment to describe the real (`:persistent_term`, BEAM-global, first-installer-wins) semantics.
- **Add `os.time_ms()` and `os.time_us()`** — non-standard extensions returning the current epoch in milliseconds / microseconds via `System.os_time/1`. Both are current-time-only (no date-table arg, since sub-second precision on a whole-second date table is meaningless). `os.time()` is unchanged.
- **`@moduledoc`** lists the new functions and explicitly marks them as extensions not present in PUC-Lua.

## Tests

Added coverage in `test/lua/vm/stdlib/os_test.exs` for integer type and magnitude of both new functions, plus a cross-check that ms/us track `os.time()`. All os tests pass (`mix test test/lua/vm/stdlib/os_test.exs` → 12 passed).

## Out of scope

- Making `os.clock()` per-VM rather than BEAM-global (issue only asks to seed at install).
- Date-table support for the `_ms`/`_us` variants.
- Changing `os.time()`'s whole-second return.

### DIFF
--- a/lib/lua/vm/stdlib/os.ex
+++ b/lib/lua/vm/stdlib/os.ex
@@ -10,6 +10,8 @@ defmodule Lua.VM.Stdlib.Os do
   ## Functions
 
   - `os.time([table])` - Epoch seconds, optionally from a date table.
+  - `os.time_ms()` - Current epoch in milliseconds (extension; not in PUC-Lua).
+  - `os.time_us()` - Current epoch in microseconds (extension; not in PUC-Lua).
   - `os.clock()` - Approximate CPU time used, in seconds.
   - `os.difftime(t2, t1)` - Difference in seconds between two times.
   - `os.date([format [, time]])` - Formats a time as a string or table.
@@ -31,6 +33,10 @@ defmodule Lua.VM.Stdlib.Os do
 
   @impl true
   def install(state) do
+    # Seed the monotonic origin at install time so os.clock() measures from a
+    # stable startup point rather than from whenever it is first called.
+    _ = boot_offset()
+
     os_table = %{
       "clock" => {:native_func, &os_clock/2},
       "date" => {:native_func, &os_date/2},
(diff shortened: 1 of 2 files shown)


lib/lua/vm/stdlib/os.ex
```diff
@@ -31,6 +33,10 @@ defmodule Lua.VM.Stdlib.Os do
 
   @impl true
   def install(state) do
+    # Seed the monotonic origin at install time so os.clock() measures from a
+    # stable startup point rather than from whenever it is first called.
+    _ = boot_offset()
```
davydog187:
```suggestion
    boot_offset()
```

PR #344, opened 2026-06-10
fix(api): deflua/2 preserves function name for guarded heads

### LINKED ISSUE #343: deflua/2 (no-state) drops the function name for guarded heads, registering it as `when`
## Summary

A `deflua` function defined **with a guard** but **without** the state/`lua` argument (i.e. the 2-arity `deflua/2` form) is registered under the name `when` instead of its real name. As a result the function is never exposed to Lua under its intended name — calling it from a script raises `attempt to call a nil value`.

## Reproduction

```elixir
defmodule MyAPI do
  use Lua.API, scope: "my"

  # guarded, no state arg -> uses deflua/2
  deflua clamp(n) when is_number(n) do
    n
  end
end

MyAPI.__lua_functions__()
#=> [{:when, false, false}]      # expected: [{:clamp, false, false}]
```

```elixir
lua = Lua.new() |> Lua.load_api(MyAPI)
Lua.eval!(lua, "return my.clamp(1)")
# ** (Lua.RuntimeException) attempt to call a nil value (field 'clamp' on global 'my')
```

The Elixir function `clamp/1` is defined correctly (it can be called directly); only its **Lua registration name** is wrong.

## Root cause

The 2-arity `deflua/2` macro extracts the name with:

```elixir
{name, _, _} = fa
```

For a guarded head the AST is `{:when, _, [{name, _, _} | _]}`, so this binds `name = :when`.

The 3-arity `deflua/3` (state) form already handles this case:

```elixir
name =
  case fa do
    {:when, _, [{name, _, _} | _]} -> name
    {name, _, _} -> name
  end
```

## Suggested fix

Use the same `case`-based name extraction in `deflua/2` that `deflua/3` already uses, so guards are supported regardless of whether the state arg is present.

## Workaround

Move the guard out of the function head and validate inside the body (or add the state arg so the `deflua/3` path is taken).

davydog187:
## Bug

`deflua/2` (a guarded head with no state arg) registered the function under the name `:when` instead of its real name, making it unreachable from Lua. For example:

```elixir
deflua clamp(a) when is_integer(a) do
  a
end
```

would register `clamp` as `:when` in `__lua_functions__/0`, so calling `clamp(...)` from a Lua chunk raised an undefined-function error.

## Root cause

In `lib/lua/api.ex`, the 2-arity `deflua` macro extracted the name with a single binding:

```elixir
{name, _, _} = fa
```

For a guarded head the AST is `{:when, _, [{name, _, _} | _]}`, so the match against the outer tuple bound `name` to `:when`. `deflua/3` (the state-arg variant) already handled this correctly with a `case`, which is why guarded heads with a state arg were never affected.

## Fix

Use the same case-based extraction `deflua/3` already uses to unwrap the `:when` node and reach the real function name, falling back to the plain head for non-guarded definitions. Name-only change; the `def unquote(fa)` body and `@variadic` handling are untouched.

## Regression test

The existing "deflua functions can have guards" test only asserted the generated Elixir function worked (`module.has_a_guard(1) == 1`) and never checked the Lua registration name, which is exactly why the bug slipped through. A new test "guarded deflua heads register under their real name, not :when" asserts:

- `__lua_functions__/0` contains the real names (`:has_a_guard`, `:with_state`) and **not** `:when`
- the guarded function is callable end-to-end from Lua via `Lua.load_api` + `Lua.eval!` (`return has_a_guard(1)`)

Confirmed the new test fails against the unfixed macro (registers `:when`) and passes with the fix.

## Validation

- `mix test test/lua/api_test.exs` — 15 passed
- `mix test` (full suite) — 2247 passed, 19 skipped, 1 excluded

Closes #343

### DIFF
--- a/lib/lua/api.ex
+++ b/lib/lua/api.ex
@@ -262,7 +262,11 @@ defmodule Lua.API do
   See `deflua/3`
   """
   defmacro deflua(fa, rest) do
-    {name, _, _} = fa
+    name =
+      case fa do
+        {:when, _, [{name, _, _} | _]} -> name
+        {name, _, _} -> name
+      end
 
     quote do
       @lua_function validate_func!(
--- a/test/lua/api_test.exs
+++ b/test/lua/api_test.exs
@@ -333,6 +333,38 @@ defmodule Lua.APITest do
       assert {1, _} = module.with_state(1, Lua.new())
       assert {"not a int", _} = module.with_state(true, Lua.new())
     end
+
+    test "guarded deflua heads register under their real name, not :when" do
+      assert [{module, _}] =
+               Code.compile_string("""
+               defmodule GuardedRegistration do
+                 use Lua.API
+
+                 deflua has_a_guard(a) when is_integer(a) do
+                   a
+                 end
+
+                 deflua has_a_guard(_) do
+                   "not a int"
+                 end
+
+                 deflua with_state(a) when is_integer(a), state do
+                   {a, state}
+                 end
+               end
+               """)
+
+      names = Enum.map(module.__lua_functions__(), &elem(&1, 0))
+
+      assert :has_a_guard in names
+      assert :with_state in names
+      refute :when in names
+
+      lua = Lua.load_api(Lua.new(), module)
+
+      assert {[1], _} = Lua.eval!(lua, "return has_a_guard(1)")
+      assert {["not a int"], _} = Lua.eval!(lua, "return has_a_guard('foo')")
+    end
   end
 
   describe "guards" do


davydog187:
## 🤖 Ultracode workflow review

Ran a 3-lens adversarial review (correctness · test-quality · regression-risk) over the diff. **Overall: approve-with-nits.**

### Findings

**nit · test-quality** — `test/lua/api_test.exs` (the `with_state` clause in the new regression test)
The new test correctly pins the bug via the `has_a_guard` clauses + `refute :when in names` + the `Lua.eval!` call-through (confirmed: it fails on unfixed `main` with `left: :when, right: [:when, :has_a_guard, :with_state]`). However the `with_state(a) when is_integer(a), state` clause registers via **`deflua/3`**, which already had the correct name extraction before this PR — so `assert :with_state in names` always passed and doesn't guard the changed path. Harmless extra coverage, not a false-positive guard.

*Optional:* drop the `with_state` clause from this test to keep it tightly scoped to the `deflua/2` path it protects. Leaving it is also fine.

### What was verified
- **Fix** mirrors the `deflua/3` case-based name extraction exactly; `deflua/2` injects no state, so the change is name-only (no prewalk needed).
- **Edge cases** all handled: multiple guarded clauses, mixed guarded/unguarded clauses, compound guards, `@variadic` + guard, zero-arity guarded heads.
- **Regression test fails without the fix** and passes with it. Full suite: 2247 passed, 19 skipped. `mix format --check-formatted` clean.

_No blocking issues._

PR #350, opened 2026-06-10
perf(table): cache array border for O(1) #t on no-hole tables

davydog187:
## What & why

Caches the array sequence border on Lua tables so `#t` is O(1) on tables with no holes, turning the per-iteration `#t` scan in `t[#t+1]=v` and `table.insert(t,v)` append loops from O(n^2) into O(n). The length operator previously rescanned the array part on every read.

## Benchmark delta

Headline: **-98%** on the default quick-mode size (n=2000 append loop, average time 21.96 ms -> 0.4449 ms). Quadratic baseline scaling (4x size -> ~16x time) collapses to linear once `#t` is O(1).

Benchee full mode, average time. Baseline `origin/main @016fa35`; branch `perf/table-cached-border @b817592` (same machine, back-to-back, detached HEAD).

**Append `t[#t+1]=v`**
| n | baseline | branch | delta |
|---|---|---|---|
| 500 | 1.33 ms (750.9 ips) | 94.49 us (10.58K ips) | -92.9% |
| 2000 | 21.96 ms (45.53 ips) | 444.86 us (2.25K ips) | -98.0% |
| 5000 | 157.44 ms (6.35 ips) | 995.73 us (1.00K ips) | -99.4% |

**`table.insert(t,v)`**
| n | baseline | branch | delta |
|---|---|---|---|
| 500 | 1.58 ms (633.9 ips) | 183.60 us (5.45K ips) | -88.4% |
| 2000 | 22.72 ms (44.01 ips) | 753.35 us (1.33K ips) | -96.7% |
| 5000 | 140.84 ms (7.10 ips) | 1.90 ms (525.8 ips) | -98.7% |

The win grows with n because the fix turns the O(n^2) per-iteration `#t` scan into O(1) — already large and clearly outside noise (baseline append deviations 9.6-13.7%, branch 6-43%) at every size.

### Reproduce

In repo, `MIX_ENV=benchmark mix deps.get` once, then a minimal Benchee script `benchmarks/table_border_append.exs` (Lua chunk eval of `for i=1,n do t[#t+1]=i end` and `for i=1,n do table.insert(t,i) end`, sizes n in {500,2000,5000}) invoked as `MIX_ENV=benchmark LUA_BENCH_MODE=full mix run benchmarks/table_border_append.exs` (full opts: time:10s warmup:2s memory_time:1s). The existing `benchmarks/table_ops.exs` does NOT cover these scenarios (its build uses `t[i]=i*i` direct indexing, never reading `#t`), so the targeted script was needed. Script was run detached and removed afterward; NOT committed to the branch.

## Review verdict

An external review (differential test against a brute-force border reference, 60k randomized ops, plus Lua-language-level confirmation) found and fixed one blocker before this PR:

- **blocker (FIXED):** Stale cached border on append-past-a-hole. The `key == arr_n+1` clause in `put_array/3` unconditionally cached `border = absorbed.arr_n` on the false premise that a freshly-written top slot proves `1..arr_n` is dense. Since `delete/2` keeps `arr_n` as a high-water mark and can leave a nil hole inside the array region, an append at `arr_n+1` resurrected a border that skips a lower hole. Repro: `t={10,20,30,40,50}; t[1]=nil; t[6]=60` gave `#t == 6` on the branch but `0` on `origin/main` and `0` from the module's own scan-based `length/1`. Both 0 and 6 are valid Lua 5.3 borders for a holey table, so not a hard spec violation, but it was a history-dependent, internally-inconsistent `#t` and an observable regression versus main that also corrupted `table.insert`/`ipairs` boundaries built on `#t`.
- **Fix (commit bf7b801):** the contiguous-append clause now only re-establishes the integer border cache when `1..arr_n` is provably dense. An integer incoming border keeps the O(1) fast path; a `:dirty` incoming border scans `array_border` once and caches only a hole-free run, otherwise stays `:dirty` so `length/1` falls back to the correct scan. Added two regression tests in `test/lua/vm/table_border_test.exs` (appending past a leading hole, appending past an interior hole).
- Remaining items were minor/nit (overloaded `:dirty` sentinel, defensive always-true branch, one-time rescan on `replace_data`-built tables) — all correctness-preserving and left as-is.

Tests pass after the fix; the reviewer's final state is **ready**.

## Merge gate

Draft, **human-gated merge** — do not merge without human sign-off.

## Independent stabilization verification (2026-06-10)

Re-verified on a clean worktree (`work-perf/table-cached-border` @ `bf7b801`) against baseline `origin/main @016fa35`. The fixes were done by another agent; this pass re-checked them skeptically.

### Blocker fix confirmed
The append-past-a-hole stale-border blocker (commit `bf7b801`) is genuinely resolved. Traced the fixed `put_array/3` contiguous-append clause: an integer incoming border re-caches in O(1) (the hot insert-loop path); a `:dirty` incoming border scans `array_border` once and only re-caches a hole-free run, otherwise stays `:dirty` so `length/1` uses the correct scan. Verified the exact repro `t={10,20,30,40,50}; t[1]=nil; t[6]=60` now reports `#t == 0` (matching `main` and the scan path), and the two pinned regression tests pass. Remaining items are doc/readability nits with no behavior impact.

### Suite
- `mix test --exclude lua53`: **2246 passed**, 7 skipped, 0 failures
- `mix test --only lua53`: **17 passed**, 12 skipped, 0 failures (`nextvar.lua` gate OK, `sort.lua` OK)
- `test/lua/vm/table_border_test.exs`: **16 passed**

### Verified numbers (interpreter-routed, n=100000)
Interpreter path forced by recursively setting `proto.bytecode = nil` (retags `:compiled_closure` -> `:lua_closure`, asserted on every benched global). `MIX_ENV=benchmark`, `LUA_BENCH_MODE=full` (time:10s, warmup:2s, memory_time:1s), Apple M4, back-to-back.

| workload | baseline (avg) | branch (avg) | delta |
|---|---|---|---|
| append `t[#t+1]=v` | ~67.8 s (1.13 min, O(n²)) | 20.82 ms | ~3258x faster (-99.97%) |
| `table.insert(t,v)` | ~67.8 s (1.13 min) | 37.34 ms | ~1816x faster (-99.95%) |
| holey `#t` loop (control) | ~22.2 ms | 23.57 ms (±1.44%) | no regression (noise) |

Memory (branch, full mode): append 194.61 MB, insert 275.46 MB, holey 245.47 MB — comparable to baseline (184.61 / 265.28 / 235.56 MB). The holey control confirms tables with holes keep `border: :dirty` and run the identical scan path on both branches, so the only change is the no-hole O(1) fast path.

**Verdict: ready** — win holds at orders of magnitude outside noise, suites green, no new regressions. Merge remains human-gated.

### DIFF
--- a/lib/lua/vm/table.ex
+++ b/lib/lua/vm/table.ex
@@ -142,11 +142,14 @@ defmodule Lua.VM.Table do
     # Cache the new border, but only if slots 1..arr_n are actually dense.
     # The append proves the top slot is filled; it says nothing about a hole
     # punched lower down (delete keeps arr_n as the high-water mark, so a
-    # cleared slot stays a nil inside the region). An integer incoming border
-    # already proves 1..(arr_n - 1) was dense, so the new run is dense too —
-    # cache in O(1), the hot insert-loop path. A :dirty incoming border might
-    # hide a lower hole, so scan once to learn the true border; a clean run
-    # re-establishes the O(1) cache, a holey one stays :dirty.
+    # cleared slot stays a nil inside the region). The load-bearing invariant:
+    # a cached integer border always equals arr_n (it is only ever set here, to
+    # absorbed.arr_n, and arr_n only grows on this same arm). So an integer
+    # incoming border proves 1..arr_n was fully dense, the new top slot extends
+    # it, and the run stays dense — cache in O(1), the hot insert-loop path. A
+    # :dirty incoming border might hide a lower hole, so scan once to learn the
+    # true border; a clean run re-establishes the O(1) cache, a holey one stays
+    # :dirty.
     case table.border do
       b when is_integer(b) ->
         %{absorbed | border: absorbed.arr_n}
@@ -240,13 +243,13 @@ defmodule Lua.VM.Table do
     # `length/1` derives the Lua border by scanning to the first hole, so a
     # cleared tail still reports the correct `#t`.
     #
-    # Cached border: punching a hole at `key` only invalidates a cached
-    # integer border `B` when `key <= B` (it could shorten the dense run
-    # reachable from 1). A hole strictly beyond `B` leaves slots `1..B`
-    # untouched and `B + 1` still absent-or-hole, so `B` stays a valid
-    # border — keep it. This keeps tail-pop loops cheap.
-    border = if dirty_after_array_hole?(table.border, key), do: :dirty, else: table.border
-    %{table | arr: :array.set(key - 1, nil, table.arr), border: border}
+    # Cached border: a cached integer border always equals arr_n (see
+    # put_array clause 1), and this clause only fires for in-array keys
+    # (1 <= key <= arr_n == border). Punching a hole anywhere in that range can
+    # only shorten the dense run reachable from 1, so every in-array delete
+    # invalidates a cached border. Drop straight to :dirty; length/1 rescans
+    # lazily on the next read.
+    %{table | arr: :array.set(key - 1, nil, table.arr), border: :dirty}
   end
 
   defp delete(%__MODULE__{data: data, dead: dead} = table, key) do
@@ -259,9 +262,6 @@ defmodule Lua.VM.Table do
       table
     end
   end
-
-  defp dirty_after_array_hole?(:dirty, _key), do: true
-  defp dirty_after_array_hole?(border, key) when is_integer(border), do: key <= border
 
   defp drop_hash_key(%__MODULE__{data: data, order: order, order_tail: tail} = table, key) do
     %{
--- a/lib/lua/vm/table.ex
+++ b/lib/lua/vm/table.ex
@@ -21,6 +21,22 @@ defmodule Lua.VM.Table do
   `data` lazily on read). A write to a sparse integer key beyond the
   border goes to `data` until a later contiguous fill promotes it.
 
+  ## Cached sequence border
+
+  `border` caches a value `length/1` is allowed to return, so the common
+  no-hole `#t` is O(1) and `table.insert` (which reads `#t` per call) stays
+  linear instead of quadratic. An integer `B` means slots `1..B` are known
+  non-nil and `B + 1` is a known hole or the array high-water end, i.e. `B`
+  is a valid Lua 5.3 §6.1 border; `length/1` returns it directly. `:dirty`
+  means "recompute": `length/1` falls back to the scan-based path.
+
+  The cache is only ever set to an integer the scan would itself return
+  (`arr_n` right after a full contiguous absorb). Every mutation that could
+  invalidate that — an in-border overwrite, a hole punched at or below the
+  cached border, a positive-integer hash insert/delete — flips it to
+  `:dirty`. Non-integer key writes never affect `#t` and leave it untouched,
+  protecting the executor's string-field fast path from churn.
+
   ## Dead-key tracking
 
   Lua 5.3 §6.1 dead-key iteration semantics are preserved for the hash
@@ -30,6 +46,7 @@ defmodule Lua.VM.Table do
 
   defstruct arr: :undefined,
             arr_n: 0,
+            border: 0,
             data: %{},
             order: [],
             order_tail: [],
@@ -41,6 +58,7 @@ defmodule Lua.VM.Table do
   @type t :: %__MODULE__{
           arr: :array.array() | :undefined,
           arr_n: non_neg_integer(),
+          border: non_neg_integer() | :dirty,
           data: %{optional(term()) => term()},
           order: list(term()),
           order_tail: list(term()),
@@ -81,6 +99,7 @@ defmodule Lua.VM.Table do
         table
         | arr: :undefined,
           arr_n: 0,
+          border: :dirty,
           data: %{},
           order: [],
           order_tail: [],
@@ -119,15 +138,22 @@ defmodule Lua.VM.Table do
   defp put_array(%__MODULE__{arr: arr, arr_n: n} = table, key, value) when key == n + 1 do
     arr = :array.set(key - 1, value, ensure_arr(arr))
     # Pull any contiguous successors that were parked in the hash into arr.
-    absorb_from_hash(%{table | arr: arr, arr_n: key})
+    # After absorb, slots 1..arr_n are dense and arr_n+1 is absent, so the
+    # resulting arr_n is a valid Lua border — cache it (O(1), no scan).
+    absorbed = absorb_from_hash(%{table | arr: arr, arr_n: key})
+    %{absorbed | border: absorbed.arr_n}
   end
 
   defp put_array(%__MODULE__{arr_n: n} = table, key, value) when key <= n do
-    %{table | arr: :array.set(key - 1, value, table.arr)}
+    # In-border overwrite. value is non-nil (put/3 routes nil to delete/2);
+    # filling a former hole could raise the reachable border, so recompute.
+    %{table | arr: :array.set(key - 1, value, table.arr), border: :dirty}
   end
 
   defp put_array(%__MODULE__{} = table, key, value) do
-    # Sparse integer beyond the border — store in hash for now.
+    # Sparse integer beyond the border — store in hash for now. insert_hash
+    # marks the border :dirty because a positive-int key could later be
+    # promoted into the sequence and extend #t.
     insert_hash(table, key, value)
   end
 
@@ -148,6 +174,11 @@ defmodule Lua.VM.Table do
   end
 
   defp insert_hash(%__MODULE__{data: data, order: order, order_tail: order_tail, dead: dead} = table, key, value) do
+    # A positive-integer hash key sits adjacent to the probe range and can
+    # change #t once a contiguous fill reaches it, so invalidate the cache.
+    # Non-integer keys (strings/floats/bools) never affect #t — leave it.
+    table = if positive_int?(key), do: %{table | border: :dirty}, else: table
+
     cond do
       Map.has_key?(dead, key) ->
         merged_order = order ++ Enum.reverse(order_tail)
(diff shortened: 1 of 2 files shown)


davydog187:

No blockers. Two test-coverage gaps in the border test suite.

- **[MINOR]** `test/lua/vm/table_border_test.exs` (whole file) — Public bulk/constructor entry points (`from_data`, `replace_data`, `put_many`) have no direct border unit test

  Every fixture is built through `put/3` via `build`/`seq`, so three public entry points that seed or rewrite the border differently are never exercised at the unit level. `from_data/1` (`lib/lua/vm/table.ex:65`) starts from the default `border: 0` and folds `put/3` over arbitrary map iteration order: a map whose keys arrive out of order (e.g. 2 before 1) must take the `:dirty` absorb-scan arm, not the O(1) integer arm — a distinct code path from the append loops the tests cover. `replace_data/1` (line 74) seeds `border: :dirty` before re-splitting. `table.sort`'s fast path (`lib/lua/vm/stdlib/table.ex:324`) writes the sorted slice back via `Table.put_many/2`, which folds `put/3` through the in-border-overwrite arm (`table.ex:108`) that flips border to `:dirty`. The invalidation logic is structurally sound and the Lua suite covers these indirectly, but no targeted unit test pins the contract for these constructors/bulk writes.

  _Suggestion:_ Add a small number of unit tests hitting the public entry points directly: `from_data` over an out-of-order/holey map (assert length and a legal border, not a stale integer), `replace_data` over a previously-dense table with an interior hole, and a `put_many` full reorder of `seq(5)` asserting length stays 5. This pins the border contract for the constructors and the sort write-back path rather than only the `put/3` append loops.

- **[NIT]** `test/lua/vm/table_border_test.exs:38-49` (`op_gen/0`) — Property generator excludes float keys and large sparse promotions — the highest-risk stale-border paths

  `op_gen` draws keys only from `integer(1..10)` and 1-2 char strings. Float-key normalization (`normalize_key` routes `t[3.0]` to the integer-3 slot, `table.ex:220`) and absorb-from-hash promotion of a large parked run are each covered by a single hand-written example, so the randomized fuzzing — the test most likely to surface an unforeseen stale-border interaction — never combines float keys or wide sparse gaps with deletes/refills. The `legal_border?` oracle would need float keys mapped to their integer slot to stay valid.

  _Suggestion:_ Add a float-key generator arm (e.g. `map(integer(1..10), &(&1 * 1.0))`) and widen the key range or add an occasional large-sparse key so the absorb/promote path is fuzzed; normalize float keys to their integer slot in the `track`/`legal_border?` oracle.

davydog187:

- **[NIT]** `test/lua/vm/table_border_test.exs:226` — Dense from_data outcome is not pinned to an integer border (O(1) re-establishment untested)

  The out-of-order from_data test (`%{2 => :b, 1 => :a, 3 => :c}`) ends dense and the code does re-establish an integer border (3) via the `:dirty` -> absorb-scan recovery arm in `put_array/3`. That arm is distinct from the plain append path and, for the from_data entry point, no test pins its outcome to an integer: the test asserts only `Table.length(t) == 3` and `legal_border?`. A regression that left this path permanently `:dirty` (correct length, lost O(1) win) would still pass. Note: the finding's reference to a "dense replace_data test" is inaccurate — the replace_data test in the suite has a holey outcome (hole at 3), so `:dirty` is correct there; only the dense from_data case is the real gap.

  Suggestion: In the out-of-order from_data test, add `assert t.border == 3` alongside the length assertion so the absorb-scan re-establishment of the O(1) cache via from_data is regression-protected.

davydog187:

- **[MINOR]** `test/lua/vm/table_border_test.exs:~150` — No regression test that non-positive integer key writes leave the cached border untouched

  `insert_hash/2` and the hash-arm `delete/2` gate border invalidation on `positive_int?(key)` (key >= 1), not `is_integer(key)`. I ran the PR code: after `seq(3)` (border=3), `Table.put(t, 0, :zero)` and `Table.put(t, -5, :neg)` both keep `border == 3` and length 3, and deleting key `0` keeps `border == 3`. The test file pins the string-key analogues ("string-field write does not dirty the border", "deleting a string key does not dirty the border") but never the non-positive-integer case, and the property generator only draws keys from `integer(1..10)` so it cannot exercise it either. A refactor swapping `positive_int?` for `is_integer` at either site would silently dirty the cache on every `t[0]`/`t[-1]` write — a real perf regression (re-scan per `#t`) with zero test signal.

  Suggested fix: add a test mirroring the string-field cases: `t = seq(3); assert Table.put(t, 0, :z).border == 3` and `Table.put(t, -5, :n).border == 3`, with `Table.length` still 3 (and a delete variant). Optionally widen `op_gen`'s key generator to include `integer(-2..0)` so the length property also covers the non-positive branch.

davydog187:

No actionable findings this round — semantics, cache-coherence, and test coverage all look sound. ✅

davydog187:

- **[MINOR]** `lib/lua/vm/table.ex:243-264` — Dead `key > border` branch in `dirty_after_array_hole?/2` + comment promises a tail-pop fast path that can never run
  - _Rationale:_ Verified the standing invariant: an integer `border` is only ever set in `put_array` clause 1 to `absorbed.arr_n`, and `arr_n` only grows on that same append arm (which always rewrites `border`). So whenever `border` is an integer it equals `arr_n` exactly. The in-array delete clause (`delete/2`, guard `key >= 1 and key <= n`) calls `dirty_after_array_hole?(table.border, key)` with `key <= arr_n == border`, so the integer arm `key <= border` is always true → always dirties. The `key > border` (keep-border) case is unreachable for in-array deletes; genuinely-beyond-border deletes route to the hash-delete clause instead. I confirmed this with a probe over append/overwrite/sparse-fill states: every integer-border state had `border == arr_n`. Not a correctness bug (it errs toward dirtying), but the comment at lines 243-247 advertises "This keeps tail-pop loops cheap" via a hole-strictly-beyond-border fast path that the cache invariant makes impossible to trigger, and the helper's general form is never exercised. Same root issue flagged separately for the `put_array` clause-1 comment (lines 142-152), which says an integer border "proves 1..(arr_n-1) was dense" when the load-bearing fact is the stronger `border == arr_n`; merged here since both are the same undocumented invariant.
  - _Suggestion:_ Either drop the helper and unconditionally set `border: :dirty` in the in-array delete clause, or keep it as defensive/future-proof code but (a) trim the lines 243-247 comment so it no longer claims a tail-pop fast path that cannot fire, and (b) in both this comment and the `put_array` clause-1 comment (142-152), state the actual load-bearing invariant: an integer border always equals `arr_n`, so `1..arr_n` was fully dense, which is why caching `arr_n` is sound and why every in-array delete must dirty.

davydog187:

- **[MINOR]** `test/lua/vm/table_border_test.exs:1-335` — No test pins border / order-memo independence after the rebase
  - _Rationale:_ Verified against the diff: the freshly-rebased order_index/order_arr memo and the new border cache now sit one line apart in the same mutation paths (insert_hash:200/210/226-227, delete:252 vs 260, drop_hash_key:272-273, replace_data:101-107). The current code invalidates them independently and correctly — I confirmed all three cases with a scratch test that passed: (a) a positive-int hash insert sets border `:dirty` AND order_index nil; (b) an in-array delete sets border `:dirty` but LEAVES order_index intact (array keys are not in `order`); (c) a string-key write keeps the integer border but nils the memo. So this is a coverage gap, not a live bug. The new test file never calls `flush_order/1`, `next_entry/2`, or asserts on order_index/order_arr, so nothing in the suite pins this independence; a future edit to these shared paths could silently leave one cache stale with zero failing tests. The PR names this interaction as its #2 risk, so the highest-risk surface of the rebase is unguarded. (Downgraded from major: the diff is correct and the existing `legal_border?`-after-every-mutation property already pins border validity under heavy interleaving — only the cross-cache independence is untested.)
  - _Suggestion:_ Add three small example tests that `flush_order/1` to populate the memo, then assert independent invalidation: (1) `seq(3) |> put("x",1) |> flush_order()` then `put(t,10,10)` => assert border == `:dirty` AND order_index == nil AND length == 3; (2) same setup then `put(t,2,nil)` (in-array delete) => assert border == `:dirty` but order_index != nil; (3) same setup then `put(t,"y",2)` (string write) => assert border stays the integer 3 while order_index == nil. These pin the exact independence the rebase merged into shared mutation paths.

davydog187:

- **[MINOR]** `test/lua/vm/table_border_test.exs:1-381` — No test pins #t correctness through table.insert/remove/sort at the Lua level for the border cache
  - _Rationale:_ The PR's headline motivation is keeping `table.insert` (which reads `#t` per call) linear. I verified at the Lua level on the PR branch that `for i=1,100 do table.insert(t,i) end`, mid-sequence `table.insert(t,pos,v)`, `table.remove(t)`/`table.remove(t,pos)`, and `table.sort` all return the correct lengths (100,101,100,99,99) — so this is a coverage gap, not a bug. The unit file pins put/put_many/from_data/replace_data directly, but the stdlib paths that actually drive the optimization (and the remove-shift border churn that repeatedly dirties/re-establishes the cache) have no regression test asserting the resulting `#t`.
  - _Suggestion:_ Add a small Lua-eval test asserting `#t` after: a `for i=1,N do table.insert(t,i) end` loop, `table.insert(t,pos,v)` mid-sequence, `table.remove(t)`/`table.remove(t,pos)`, and `table.sort`. This pins the headline insert-loop path and the remove-shift border churn the put-level tests only approximate.

- **[MINOR]** `test/lua/vm/table_border_test.exs:297-340` — No test asserts the order memo is dropped (not left stale) when absorb promotes a parked key out of the hash
  - _Rationale:_ Round-3 RISK (2) is that border and order_index/order_arr must invalidate independently and correctly in the shared mutation paths. The added independence tests cover positive-int hash insert (drops both), in-array delete (drops border, keeps memo), and string write (keeps border, drops memo) — but NOT the absorb path. I confirmed on the PR branch that `absorb_from_hash` calls `drop_hash_key`, which clears order_index/order_arr (and re-establishes an integer border), so the code is correct. But the sequence `seq(2); t[4]=4; flush_order; t[3]=3` (key 4 promoted out of the hash via absorb) is unpinned: nothing asserts the memo is nilled AND that `pairs()` does not surface a stale hash entry for the now-promoted key 4.
  - _Suggestion:_ Add a test: build `seq(2)`, park key 4 in the hash, `flush_order` to populate the memo, then put key 3 to trigger absorb of 4; assert `border == 4`, `order_index == nil`, `order_arr == nil`, and that `pairs()` yields keys 1..4 from the array each exactly once with no duplicate/stale entry for 4.

- **[NIT]** `test/lua/vm/table_border_test.exs:29-40` — Property generator never produces integer-valued float keys, leaving the float/border collapse thinly covered
  - _Rationale:_ `op_gen` draws keys only from `integer(1..10)`, `integer(-2..0)`, and short strings. A float like `2.0` normalizes via `normalize_key/1` (in `put/3`, before any border logic) to integer slot 2, so it can extend/punch the border exactly like integer 2. The property never drives this collapse interleaved with deletes/refills; only the single happy-path example `float keys behave like their integer slots` covers it. Coverage value is incremental — floats are normalized to integers before the border layer ever sees them, so the integer branches the property already hammers exercise the same code — but the fold is cheap and widens the oracle's reach.
  - _Suggestion:_ Add an integer-valued-float branch to `op_gen`, e.g. `{2, map(integer(1..10), &(&1 * 1.0))}`, so the property exercises float-to-integer key collapse against the border and the `get/2` oracle. Optionally add one example where a float key fills a hole below the cached border and asserts the resulting length.

- **[NIT]** `lib/lua/vm/table.ex:28-30` — Moduledoc claims integer border B+1 is 'a known hole or the array high-water end' but B always equals arr_n, so B+1 is only ever the high-water end
  - _Rationale:_ Verified on the PR branch: border is only ever assigned `%{absorbed | border: absorbed.arr_n}` (the two put_array arms), and any hole at or below arr_n forces `:dirty`. So an integer border always equals `arr_n` and B+1 is strictly the first unallocated slot, never an in-region hole. The 'known hole' half of the disjunction describes a state the code never reaches, slightly overstating what the cache represents versus the load-bearing invariant already spelled out in the put_array clause-1 comment.
  - _Suggestion:_ Tighten the wording to the invariant the code maintains: an integer border B always equals `arr_n`, slots 1..B are dense, and B+1 is the array high-water end (any hole at or below arr_n forces `:dirty`).

PR #352, opened 2026-06-10
perf(vm): rebuild call_stack lazily on the Lua->Lua call path

davydog187:
## What changed and why

Rebuilds `state.call_stack` lazily at native-call boundaries instead of materializing call_info + tuple frames on every Lua->Lua call. The hot recursive Lua path drops two list ops and a 4-key map allocation per call; the cost moves to the (comparatively rare) native boundary via `rebuild_call_stack/1`. Adds `call_iterator/6` to cover the `generic_for` materialization gap.

## Benchmark — interpreter path (the real signal)

The original fib(30) measurement below routed through the **dispatcher**, which does not exercise this branch's changed code path, so it showed nothing. Re-measured on the **tree-walking interpreter** (fib stripped to `:lua_closure` via the `benchmarks/dispatcher_vs_interpreter.exs` bytecode-strip mechanism), the allocation reduction is clear and deterministic:

| ref | fib(28) ips | fib(28) mem/iter | fib(30) mem/iter |
|-----|------------|------------------|------------------|
| baseline `016fa35` | 2.79 / 2.68 | 2.84 GB | 7.43 GB |
| this branch | 2.83 / 2.81 | **2.78 GB (-2.1%)** | **7.29 GB (-1.9%)** |

- **Memory per iteration: -2% on the hot Lua->Lua path, deterministic (0.00% deviation across runs)** — exactly this branch's intent (no per-call `call_info` map). This is the noise-immune proof.
- **Wall-clock**: this branch was faster in *both* repeat rounds (~1.5-5%), tracking the allocation win rather than thermal luck. Modest but consistently in-favor.

Config: Benchee `time:10 warmup:2 memory_time:3`, sequential runs, 2 repeats per ref, fib forced through `Lua.VM.Executor`, correctness asserted (fib(28)==317811, fib(30)==832040). Bench script not committed.

### Original (dispatcher-path) measurement — superseded, retained for the record

Scenario: `fib(30)` recursive via pre-compiled chunk. delta_pct **+0.4%**, INCONCLUSIVE — this routed through the dispatcher and never touched the changed call path. Superseded by the interpreter-path numbers above.

## Review

Motivated by an external code review of the perf branch. Verdict: **ready** (tests pass after; no fixes required). Two nits, both non-blocking:

1. **nit** — The `__call` metamethod dispatch path (`executor.ex` ~line 1237) invokes `call_function(call_mm, ...)` without materializing the lazily-tracked Lua frames into `state.call_stack`, so a `__call` callback that reads `debug.traceback`/`getinfo` or raises sees a truncated stack (missing in-flight executor frames). Differential testing against `origin/main` confirmed this is PRE-EXISTING behavior, byte-identical on both branches, NOT a regression. The author added `call_iterator/6` to cover the analogous `generic_for` gap, but the `__call` path has the same gap on main too; out of scope for this perf change. Worth a follow-up but not a blocker.
2. **nit** — `rebuild_call_stack/1` is O(depth) and is invoked at every native-call boundary (and prepended via `++` at three error raise sites). For native-call-heavy workloads at deep stacks this is more per-native-call work than the old design (which only read the already-materialized list). The net win still holds because the hot Lua->Lua path drops two list ops + a 4-key map alloc per call, and native calls are comparatively rare; but the cost moved from the common path to the native boundary rather than disappearing. Recommend running the run-bench harness on a call-heavy + debug-heavy workload to confirm no regression at the native boundary.

## Merge gating

Human-gated merge. Marked ready for review; do NOT merge automatically — merge on maintainer go-ahead after reviewing the diff and the interpreter-path numbers above.

### Final verified numbers (interpreter path, bytecode-stripped, `:lua_closure` asserted; `MIX_ENV=benchmark LUA_BENCH_MODE=full`, Apple M4)

**Allocation win (holds, independently reproduced vs main @016fa35):**

| workload | main alloc | branch alloc | delta | main time | branch time |
|---|---|---|---|---|---|
| fib(28) | 2.84 GB | 2.78 GB | **-2.1%** | 380 ms | 359 ms |
| fib(30) | 7.43 GB | 7.29 GB | **-1.9%** | 997 ms | 979 ms |

Memory deviation ±0.00% on both (deterministic).

> Note: the perf claim was re-measured against a freshly built `016fa35`. An earlier "does-not-reproduce" reading was a baseline mismeasurement (the reported main figures were actually the branch's). A clean main build lands at 2.84 / 7.43 GB.

**Native-boundary trade-off (documented, accepted):** `debug.traceback` x50 at depth pays for the lazy approach — `rebuild_call_stack/1` is O(depth) on demand.

| workload | main | branch |
|---|---|---|
| traceback x50 depth40 | 148 µs / 595 KB | 203 µs / 730 KB |
| traceback x50 depth60 | 230 µs / 826 KB | 302 µs / 1030 KB |

This native-boundary cost funds the hot-path allocation win on the Lua→Lua call path.

### `__call` boundary fix (commit cbf7068)
Materializes the lazily-tracked executor `frames` into `state.call_stack` at the `__call` metamethod dispatch site, mirroring the native-dispatch / dispatcher / generic_for boundaries, and adds `__call` traceback/getinfo/balance coverage to `lazy_call_info_test.exs`.

Independent verification note: differential probing of the pre-fix commit found the `__call` path already byte-identical to main (`debug.traceback` line 7, `debug.getinfo(2,'nl').currentline` 7, including a 4-deep `__call` chain), so this change is **behavior-neutral / defensive** rather than correcting an observed divergence. It is harmless and keeps the lazy bookkeeping balanced across all stack-reading boundaries.

### Suite
Full unit: 2259 passed, 19 skipped, 1 excluded. lua53 (`--only lua53`, errors/db guardrails): 17 passed, 12 skipped. `lazy_call_info_test.exs`: 12 passed. `mix format` clean.

### DIFF
--- a/lib/lua/vm/executor.ex
+++ b/lib/lua/vm/executor.ex
@@ -1074,7 +1074,6 @@ defmodule Lua.VM.Executor do
         # without going through this branch.
         args = collect_args(regs, base + 1, total_args)
         call_info = %{source: proto.source, line: line, name: hint_name(name_hint), namewhat: hint_namewhat(name_hint)}
-        State.check_call_depth!(state)
 
         # The executor's own in-flight Lua frames are tracked lazily via the
         # `frames` argument and are NOT in `state.call_stack`. Materialize them
@@ -1083,10 +1082,15 @@ defmodule Lua.VM.Executor do
         # `debug.*` / tracebacks. Restore the inherited stack on return so the
         # lazy bookkeeping stays balanced.
         inherited_call_stack = state.call_stack
+        materialized_call_stack = [call_info | rebuild_call_stack(frames) ++ inherited_call_stack]
+
+        # Check depth against the materialized stack so an overflow here
+        # reports the full traceback, not the bare inherited stack.
+        State.check_call_depth!(state, materialized_call_stack)
 
         state = %{
           state
-          | call_stack: [call_info | rebuild_call_stack(frames) ++ inherited_call_stack],
+          | call_stack: materialized_call_stack,
             call_depth: state.call_depth + 1
         }
 
@@ -1133,7 +1137,18 @@ defmodule Lua.VM.Executor do
           name_hint: name_hint
         }
 
-        State.check_call_depth!(state)
+        # Hot path: `call_depth_ok?/1` is a pure comparison, so the lazy
+        # frames are only materialized into a traceback when the limit is
+        # actually hit. `frames` excludes the call about to be made, so its
+        # synthesized entry is prepended.
+        if State.call_depth_ok?(state) do
+          :ok
+        else
+          overflow_call_stack =
+            [dispatcher_call_info(proto, name_hint, line) | rebuild_call_stack(frames)] ++ state.call_stack
+
+          State.check_call_depth!(state, overflow_call_stack)
+        end
 
         state = %{
           state
--- a/lib/lua/vm/executor.ex
+++ b/lib/lua/vm/executor.ex
@@ -351,41 +351,80 @@ defmodule Lua.VM.Executor do
   def dispatcher_bitwise(:band, a, b, state, proto, hint_a, hint_b) do
     src = proto.source
 
-    try_binary_metamethod("__band", a, b, state, fn ->
-      Numeric.to_signed_int64(Bitwise.band(to_integer!(a, 0, src, hint_a, state), to_integer!(b, 0, src, hint_b, state)))
-    end)
+    try_binary_metamethod(
+      "__band",
+      a,
+      b,
+      state,
+      fn ->
+        Numeric.to_signed_int64(
+          Bitwise.band(to_integer!(a, 0, src, hint_a, state), to_integer!(b, 0, src, hint_b, state))
+        )
+      end,
+      :materialized
+    )
   end
 
   def dispatcher_bitwise(:bor, a, b, state, proto, hint_a, hint_b) do
     src = proto.source
 
-    try_binary_metamethod("__bor", a, b, state, fn ->
-      Numeric.to_signed_int64(Bitwise.bor(to_integer!(a, 0, src, hint_a, state), to_integer!(b, 0, src, hint_b, state)))
-    end)
+    try_binary_metamethod(
+      "__bor",
+      a,
+      b,
+      state,
+      fn ->
+        Numeric.to_signed_int64(Bitwise.bor(to_integer!(a, 0, src, hint_a, state), to_integer!(b, 0, src, hint_b, state)))
+      end,
+      :materialized
+    )
   end
 
   def dispatcher_bitwise(:bxor, a, b, state, proto, hint_a, hint_b) do
     src = proto.source
 
-    try_binary_metamethod("__bxor", a, b, state, fn ->
-      Numeric.to_signed_int64(Bitwise.bxor(to_integer!(a, 0, src, hint_a, state), to_integer!(b, 0, src, hint_b, state)))
-    end)
+    try_binary_metamethod(
+      "__bxor",
+      a,
+      b,
+      state,
+      fn ->
+        Numeric.to_signed_int64(
+          Bitwise.bxor(to_integer!(a, 0, src, hint_a, state), to_integer!(b, 0, src, hint_b, state))
+        )
+      end,
+      :materialized
+    )
   end
 
   def dispatcher_bitwise(:shl, a, b, state, proto, hint_a, hint_b) do
     src = proto.source
 
-    try_binary_metamethod("__shl", a, b, state, fn ->
-      lua_shift_left(to_integer!(a, 0, src, hint_a, state), to_integer!(b, 0, src, hint_b, state))
-    end)
+    try_binary_metamethod(
+      "__shl",
+      a,
+      b,
+      state,
+      fn ->
+        lua_shift_left(to_integer!(a, 0, src, hint_a, state), to_integer!(b, 0, src, hint_b, state))
+      end,
+      :materialized
+    )
   end
 
   def dispatcher_bitwise(:shr, a, b, state, proto, hint_a, hint_b) do
     src = proto.source
 
-    try_binary_metamethod("__shr", a, b, state, fn ->
-      lua_shift_right(to_integer!(a, 0, src, hint_a, state), to_integer!(b, 0, src, hint_b, state))
-    end)
+    try_binary_metamethod(
+      "__shr",
+      a,
+      b,
+      state,
+      fn ->
+        lua_shift_right(to_integer!(a, 0, src, hint_a, state), to_integer!(b, 0, src, hint_b, state))
+      end,
+      :materialized
+    )
   end
 
   @doc false
@@ -393,9 +432,15 @@ defmodule Lua.VM.Executor do
   def dispatcher_bnot(val, state, proto, hint) do
     src = proto.source
 
-    try_unary_metamethod("__bnot", val, state, fn ->
-      Numeric.to_signed_int64(Bitwise.bnot(to_integer!(val, 0, src, hint, state)))
-    end)
+    try_unary_metamethod(
+      "__bnot",
+      val,
+      state,
+      fn ->
+        Numeric.to_signed_int64(Bitwise.bnot(to_integer!(val, 0, src, hint, state)))
+      end,
+      :materialized
+    )
   end
 
   @doc false
--- a/lib/lua/vm/executor.ex
+++ b/lib/lua/vm/executor.ex
@@ -243,79 +243,128 @@ defmodule Lua.VM.Executor do
   @spec dispatcher_binop(atom(), term(), term(), State.t(), term(), term(), term()) ::
           {term(), State.t()}
   def dispatcher_binop(:add, a, b, state, proto, hint_a, hint_b) do
-    try_binary_metamethod("__add", a, b, state, fn ->
-      safe_add(a, b, 0, proto.source, hint_a, hint_b, state)
-    end)
+    try_binary_metamethod(
+      "__add",
+      a,
+      b,
+      state,
+      fn ->
+        safe_add(a, b, 0, proto.source, hint_a, hint_b, state)
+      end,
+      :materialized
+    )
   end
 
   def dispatcher_binop(:subtract, a, b, state, proto, hint_a, hint_b) do
-    try_binary_metamethod("__sub", a, b, state, fn ->
-      safe_subtract(a, b, 0, proto.source, hint_a, hint_b, state)
-    end)
+    try_binary_metamethod(
+      "__sub",
+      a,
+      b,
+      state,
+      fn ->
+        safe_subtract(a, b, 0, proto.source, hint_a, hint_b, state)
+      end,
+      :materialized
+    )
   end
 
   def dispatcher_binop(:multiply, a, b, state, proto, hint_a, hint_b) do
-    try_binary_metamethod("__mul", a, b, state, fn ->
-      safe_multiply(a, b, 0, proto.source, hint_a, hint_b, state)
-    end)
+    try_binary_metamethod(
+      "__mul",
+      a,
+      b,
+      state,
+      fn ->
+        safe_multiply(a, b, 0, proto.source, hint_a, hint_b, state)
+      end,
+      :materialized
+    )
   end
 
   def dispatcher_binop(:divide, a, b, state, proto, hint_a, hint_b) do
-    try_binary_metamethod("__div", a, b, state, fn ->
-      safe_divide(a, b, 0, proto.source, hint_a, hint_b, state)
-    end)
+    try_binary_metamethod(
+      "__div",
+      a,
+      b,
+      state,
+      fn ->
+        safe_divide(a, b, 0, proto.source, hint_a, hint_b, state)
+      end,
+      :materialized
+    )
   end
 
   def dispatcher_binop(:floor_divide, a, b, state, proto, hint_a, hint_b) do
-    try_binary_metamethod("__idiv", a, b, state, fn ->
-      safe_floor_divide(a, b, 0, proto.source, hint_a, hint_b, state)
-    end)
+    try_binary_metamethod(
+      "__idiv",
+      a,
+      b,
+      state,
+      fn ->
+        safe_floor_divide(a, b, 0, proto.source, hint_a, hint_b, state)
+      end,
+      :materialized
+    )
   end
 
   def dispatcher_binop(:modulo, a, b, state, proto, hint_a, hint_b) do
-    try_binary_metamethod("__mod", a, b, state, fn ->
-      safe_modulo(a, b, 0, proto.source, hint_a, hint_b, state)
-    end)
+    try_binary_metamethod(
+      "__mod",
+      a,
+      b,
+      state,
+      fn ->
+        safe_modulo(a, b, 0, proto.source, hint_a, hint_b, state)
+      end,
+      :materialized
+    )
   end
 
   def dispatcher_binop(:power, a, b, state, proto, hint_a, hint_b) do
-    try_binary_metamethod("__pow", a, b, state, fn ->
-      safe_power(a, b, 0, proto.source, hint_a, hint_b, state)
-    end)
+    try_binary_metamethod(
+      "__pow",
+      a,
+      b,
+      state,
+      fn ->
+        safe_power(a, b, 0, proto.source, hint_a, hint_b, state)
+      end,
+      :materialized
+    )
   end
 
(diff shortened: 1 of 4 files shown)


davydog187:

- **[MAJOR]** `lib/lua/vm/executor.ex:1136` (also `:1077`) — Stack-overflow traceback is empty on the lazy `:lua_closure` call path

  Confirmed by direct reproduction against the PR head (`cbf7068`). `State.check_call_depth!/1` raises the "stack overflow" error with `call_stack: state.call_stack` (`state.ex:82`). The PR stops pushing per-call frames on the `:lua_closure` path (the frames now live only in the lazy `frames` argument), so when `check_call_depth!` fires at `executor.ex:1136` `state.call_stack` is empty. A 50-deep recursion forced down the pure-interpreter `:lua_closure` path (bytecode stripped — exactly the bytecode-fallback case from `compiler/bytecode.ex:370`) rendered `call_stack length: 0` and emitted **no** "Stack trace:" section, raised from `executor.ex:1136`. The identical recursion through the normal compiled/dispatcher path rendered the full 50-frame traceback. The `:compiled_closure` branch shares the defect: `check_call_depth!` at line `1077` runs before the frames are materialized into `state.call_stack` at lines `1085-1091`. The error still carries the correct message and value, so this is a traceback-fidelity regression, not a wrong result — but it is reachable by real Lua whenever a recursive function body contains any bytecode-unsupported instruction. The PR's own overflow test only matches `/stack overflow/` and never inspects the traceback, so it does not catch this.

  *Suggestion:* Materialize the lazy frames into the stack used for the overflow report at both depth-check sites. E.g. at line `1136` build `overflow_stack = rebuild_call_stack(frames) ++ state.call_stack` and have the depth check raise with that (pass it to a `check_call_depth!` arity that accepts an explicit stack); at line `1077` move/replicate the materialization before the check so the compiled-closure overflow also includes the rebuilt frames. Add a regression test on the pure-interpreter path that asserts the traceback contents (frame count / an `in function 'rec'` line), not just the message regex.

davydog187:

- **[MAJOR]** `test/lua/vm/lazy_call_info_test.exs:1-216` — lazy_call_info_test.exs exercises the `:compiled_closure` dispatcher branch, not the lazy `:lua_closure` path it documents

  The moduledoc claims these tests guard the lazily-rebuilt call_stack on the executor's Lua→Lua path where `state.call_stack` is left untouched — that is the `:lua_closure` clause at `executor.ex:1101`, the branch this PR optimizes (frame now carries `:line`/`:name_hint`). But every program here uses plain `function f()...end`/`setmetatable` closures, and those sub-prototypes are emitted *with* bytecode (`[true, true]`) per `Compiler.compile`. Per `executor.ex:1029-1033` a non-nil-bytecode closure is tagged `:compiled_closure`, so nested calls route through the `:compiled_closure` hand-off at `executor.ex:1068` — a different materialization site. Proof: the suite asserts `debug.getinfo` `currentline == 0` (line 48, compiled path strips line info), whereas forcing the genuine lazy path with a bitwise op (`3 & 1`, which can't be bytecode-encoded) yields `currentline == 3`. So getinfo/traceback/`__call` fidelity *and* the call_stack-balance assertions on the real lazy `:lua_closure` branch are unguarded. The path is correct today (forced-lazy getinfo returns `["outer","global",3,"test.lua","Lua"]` and call_stack balances to `[]`), so this is a test-coverage gap rather than a live bug. `recursion_depth_test.exs` already uses the bitwise trick to cover the lazy *overflow* traceback, so that sub-case is genuinely guarded; the gap is the getinfo/traceback/`__call`/native-callback assertions. A future change to `rebuild_call_stack/1` or the frame shape could break the lazy path with every test in this file still green.

  *Suggestion:* Add the bitwise-op trick (as in `recursion_depth_test.exs`'s `@lazy_recursion`) to each scenario so outer/inner/the `__call` callback/the native-callback `deep`/`outer` (lines 169-183) fall onto the `:lua_closure` path, OR add parallel lazy-variant tests. At minimum cover on the forced-lazy path: `debug.getinfo(2,"nSl")` fields for a nested chain, `debug.traceback` line content, the `__call` callback currentline/traceback, the native-dispatch save/restore (with `state.call_stack == []` post-condition), and call_stack balance. Update the moduledoc to match what is actually exercised.

- **[MINOR]** `lib/lua/vm/executor.ex:1085-1089` — `:compiled_closure` hand-off materializes the full call_stack on every success, contradicting the `rebuild_call_stack/1` "never on the hot success path" doc

  Verified against PR head `9a50814`. The `:compiled_closure` branch builds `materialized_call_stack = [call_info | rebuild_call_stack(frames) ++ inherited_call_stack]` unconditionally on every interpreter→dispatcher hand-off (lines 1085-1089), then passes it to `check_call_depth!/2`. The parallel `:lua_closure` branch (lines 1144-1151) instead gates `rebuild_call_stack(frames)` behind `State.call_depth_ok?/1`, so the O(depth) `Enum.map` walk only runs on overflow. Meanwhile `rebuild_call_stack/1`'s own doc comment (`executor.ex:555-557`) states it is "Called only at the boundaries that actually read the stack ... never on the hot success path," and the frame-building comment claims lazy frames are "only materialized into a traceback when the limit is actually hit." Both statements are inaccurate for the `:compiled_closure` clause, which pays the walk eagerly on every hand-off success. Not a correctness bug — the stack is restored to `inherited_call_stack` on return (line 1098) and the suite is green — but it is a real per-hand-off cost the doc says is avoided.

  *Suggestion:* Either gate the materialization behind `State.call_depth_ok?/1` the way the `:lua_closure` clause does (only building `materialized_call_stack` when actually needed), OR — if the dispatcher genuinely needs a live stack for its whole duration because it can run native callbacks mid-execution — tighten the `rebuild_call_stack/1` doc and the frame-building comment to acknowledge that the interpreter→dispatcher boundary materializes eagerly, so the "never on the hot success path" wording is not contradicted.

davydog187:

### Blockers

- **[BLOCKER]** `lib/lua/vm/executor.ex:2429` (index_value `__index`), `:2520` (table_index `__index`), `:2577` (table_newindex `__newindex`), `:2707/2711/2715` (invoke_metamethod) — Metamethod dispatch (`__index`/`__newindex`/arithmetic/concat/`__len`/comparison) drops enclosing Lua frames from `state.call_stack` on the lazy path
  - Empirically confirmed regression, not speculation. The PR materializes lazy `frames` into `state.call_stack` at the `__call` boundary (line 1247), native dispatch (1176), the dispatcher hand-off (1090), and `call_iterator/6` — but the other metamethod dispatch helpers that re-enter Lua via `call_function/3` were missed. `index_value/6` (2429), `table_index` (2520), `table_newindex` (2577), and `invoke_metamethod/4` (2707/2711/2715) are all invoked from inside `do_execute` opcode clauses (e.g. `index_value` from 1859/1885/1892/1959/1966/2047) where `frames` is in scope and holds the live executor stack, but none thread `frames` in or materialize it. `call_function/3` then recurses with `frames=[]` (executor.ex:157), so a metamethod that reads `state.call_stack` sees a truncated stack. Reproduced on the PR head (9a50814): a `__index` metamethod calling `debug.traceback('idx')` from a forced-lazy Lua function returns `"idx\nstack traceback:"` (enclosing `driver` frame GONE) and `debug.getinfo(2,'nl').currentline` returns -1. On merge-base 016fa35 the same program returns `"idx\nstack traceback:\n\ttest.lua:9: in ?"` and `getinfo(2)` returns 9. This breaks Lua 5.3 debug semantics and corrupts error tracebacks raised from inside metamethods on the lazy interpreter path.
  - **Suggestion:** Thread `frames` into `index_value/6`, `table_index`, `table_newindex`, and `invoke_metamethod/4` and wrap the `call_function/3` calls at 2429/2520/2577/2707-2715 with the same inherited-stack/`rebuild_call_stack(frames)`/restore dance already used for `__call` at 1247. Add regression tests asserting `debug.traceback` / `debug.getinfo` / error tracebacks inside `__index`, `__newindex`, and an arithmetic metamethod still observe the enclosing Lua frame, on both the compiled and forced-lazy paths.

### Minor

- **[MINOR]** `test/lua/vm/lazy_call_info_test.exs:1-377` — No regression test for the generic_for iterator materialization boundary (`call_iterator/6`)
  - The diff introduces `call_iterator/6` (executor.ex:582-588) specifically to materialize lazy `frames` into `state.call_stack` for a generic_for iterator call, and rewires both for-loop call sites to it (executor.ex ~689 and ~974). This is one of the five read boundaries the PR's own moduledoc enumerates, yet no test in the new file (or recursion_depth_test.exs) drives a `for ... in` loop combined with `debug.*` — confirmed by grep returning nothing. The path is currently correct, so this is a missing regression test rather than a live bug; but finding #1 demonstrates that exactly these untested materialization boundaries are where the lazy refactor silently truncates tracebacks.
  - **Suggestion:** Add a describe block driving `for _, v in iter, nil, 0 do` where `iter` calls `debug.traceback` / `debug.getinfo(2,'nl').currentline`, assert the enclosing frame's call-site line appears, cover both compiled and forced-lazy variants, and assert `call_stack` restores to `[]` afterward to pin the inherited-stack restore in `call_iterator/6`.

### Nits

- **[NIT]** `test/lua/vm/lazy_call_info_test.exs:68-78` (and every forced-lazy test) — Lazy-path tests force the interpreter via the implicit `3 & 1` codegen quirk instead of the existing `strip_bytecode/1` helper
  - Every forced-lazy test sprinkles `local _ = 3 & 1` to force the `:lua_closure` path, relying on bitwise ops being non-bytecode-encodable. If a future codegen change makes `&` encodable, these tests silently revert to the compiled path and stop exercising the lazy branch they are named for, with no failing signal. The repo already has an intent-revealing alternative in `pcall_state_preservation_test.exs:44-48` (`strip_bytecode/1` recursively nils `proto.bytecode`) run under `for engine <- [:compiled, :interpreted]`, which forces the interpreter path structurally and independent of opcode coverage.
  - **Suggestion:** Adopt the `strip_bytecode/1` + per-engine `for engine <- [:compiled, :interpreted]` pattern from `pcall_state_preservation_test.exs` so the lazy path is forced structurally, making coverage robust to codegen changes and removing the duplicate golden line numbers per variant.

davydog187:

- **[NIT]** `test/lua/vm/lazy_call_info_test.exs:262-296` — No regression test for a native-raised error's traceback relying on the line-1197 stack materialization

  The native-dispatch boundary materializes the lazy executor frames into `state.call_stack` at `executor.ex:1197` (`rebuild_call_stack(frames) ++ inherited`). On the error-from-native path, that materialized state is attached to the raised exception via `annotate_frame_state/2` (`executor.ex:1213`, fills `:state` only when nil). This is the sole source of traceback fidelity for that path: a forced-lazy chain `outer -> inner -> ("x"):rep(true)` raises with the exception's OWN `:call_stack` empty (length 0) while the attached `:state.call_stack` carries the 2 outer frames. This boundary is distinct from both the success-path getinfo/traceback reads (which read the live stack before any raise) and the executor's own raise sites at `executor.ex:1224/1236/1248` (which use the error's own `:call_stack` and ARE covered by the "attempt to call a nil value" tests). A regression in the line-1197 materialization would silently truncate stdlib-error tracebacks with no failing test. The code is correct today; this is purely a coverage gap on a perf PR whose stated risk is exactly "a consumer that reads the stack observes a truncated stack".

  _Suggestion:_ Add a small regression: a forced-lazy nested chain (`outer -> inner`) where `inner` triggers a native stdlib arg error (e.g. `("x"):rep(true)`), and assert the rendered traceback / attached `state.call_stack` contains the outer frame. This pins the native-boundary materialization on the error path, which the existing success-path getinfo tests do not exercise.

davydog187:
🤖 **Automated CI status update**

The CI failure displayed on this PR (run [27343957506](https://github.com/tv-labs/lua/actions/runs/27343957506)) is **stale**. It ran against an earlier commit that had a compile error (`try_binary_metamethod`/`try_unary_metamethod` arity mismatch). That has already been fixed by the current branch tip **42578fd** ("fix(vm): materialize lazy frames at metamethod dispatch boundaries").

I independently verified the current tip locally:

- `mix compile` — succeeds ("Generated lua app")
- `mix format --check-formatted` — exit 0
- `mix test` — `Result: 2280 passed (62 doctests, 51 properties, 2167 tests), 19 skipped, 1 excluded`

CI has been re-triggered (`gh run rerun 27343957506 --failed`) so the PR reflects the green tip. No code changes were made.

davydog187:
### ⚠️ Correction to my earlier automated comment

My previous comment said the failure was "stale" and the tip was already green — **that was wrong**, and I apologize for the noise. Re-triggering the old run kept failing because `gh run rerun` re-checks-out the run's *original* commit, and more importantly the branch genuinely did not build once combined with `main`.

### Real diagnosis

CI builds `refs/pull/352/merge` — this PR's head **merged into `main`** — not the branch tip in isolation. The branch compiles alone, but the **merge with `main` does not**. Two semantic conflicts with `main`'s #347 (`perf(vm): compile bitwise ops`):

1. **Compile error.** #347 added `dispatcher_bitwise/7` + `dispatcher_bnot/4`, which call `try_binary_metamethod`/`try_unary_metamethod` at the pre-lazy arity. This branch added a trailing `frames` arg to those helpers, so the merged tree failed with `try_binary_metamethod/5 undefined`. Fixed by threading `:materialized` through the six bitwise/bnot call sites (matching the sibling `dispatcher_*` helpers).

2. **Silent coverage loss.** #347 also made bitwise ops bytecode-encodable, so the `3 & 1` token `lazy_call_info_test.exs` used to force the `:lua_closure` interpreter path now *compiles* — quietly moving every forced-lazy case onto the compiled path (line attribution `0`/`-1` instead of the live line). Three line-sensitive cases failed; the rest passed while no longer testing the lazy path. Switched the forcing op to `1 and 2` (its `test_and` opcode is still non-encodable), restoring genuine lazy-path coverage with the line goldens intact.

### What I pushed

- Merged `origin/main` into the branch.
- Applied both fixes above (commit `43ba9fb`).
- **Verified locally on the merged tree:** `mix compile` clean, `mix format --check-formatted` clean, full suite **2373 passed, 19 skipped**.

CI is re-running against the new tip now.

_Automated message from Claude Code._

davydog187:
### Resolved merge conflicts with `main` (post-#355)

Now that #355 is merged, `main` and this branch both reworked the same dispatcher functions in `lib/lua/vm/executor.ex`, producing 10 conflict hunks. Both features had to be combined, not chosen between:

- **#355** changed non-call dispatcher line attribution from `0` → `nil` (so the error render falls back to `current_position/0` instead of printing `:0:`).
- **#352** added the lazy-frames marker (`:materialized` / the extra `frames` arg) to those same call sites.

Resolution = **main's `nil` line semantics + #352's `:materialized` threading** across `dispatcher_binop`, `dispatcher_unop`, `dispatcher_cmp`, `dispatcher_get_field`/`get_table`/`index_method_target`, and `dispatcher_concat`.

One golden also moved: the compiled-path `level 1 currentline` test asserted `-1` (line stripped). #355 now bakes the per-call line into the call opcode, so the live frame resolves to its call site (`2`). Updated the golden + moduledoc note (commit `43b4adb`).

**Verified locally on the merged tree:** `mix compile` clean, `mix format --check-formatted` clean, full suite **2384 passed, 19 skipped**.

_Automated message from Claude Code._

davydog187:
> The case against rushing it: #352 optimizes the tree-walking interpreter (:lua_closure) path — about −2% allocations on hot Lua→Lua recursion. But that path is only taken when a function can't be lowered to bytecode, and the project is actively moving the opposite direction: #347 just made bitwise ops compile, and more compiled-path work is ongoing. So the win targets a shrinking fraction of real execution, while the default hot path is the dispatcher, which this PR doesn't speed up.
> 
> The real cost is maintenance, not the runtime trade-off. The PR's own benchmarks already disclose the native-boundary regression (deep + debug-heavy workloads pay for rebuild_call_stack/1 being O(depth)) and that's an acceptable, documented trade. The bigger thing is structural: every site that reads state.call_stack must now remember to materialize the lazy frames first. This very branch needed a string of follow-up commits patching boundaries (__call, stack-overflow traceback, metamethod dispatch), and this merge surfaced yet another interaction. It's a standing footgun — any future code that reads the call stack and forgets to materialize gets a silently truncated stack.
> 
> My recommendation: I'd lean toward deferring/declining unless the tree-walking interpreter path genuinely matters for your workloads, because you're paying a permanent correctness-maintenance tax for a modest win on a path you're trying to make less common. If you do want it, I'd at least want the "materialize before reading call_stack" invariant documented in the executor (a comment at each boundary + the helper) so the footgun is visible to the next person.
> 
> Either way it's ready — green and mergeable — if you decide the interpreter-path win is worth it.

PR #364, opened 2026-06-14
feat(vm): conformant goto/label on both engines

davydog187:
## Context

Second PR in the effort to drive the bytecode **Dispatcher** to full opcode coverage (follows #362). The remaining real gap was `:goto` / `:label`.

While implementing the dispatcher port, I found the premise was wrong: the **interpreter's `goto` was itself non-conformant**. It resolved labels with a forward-only `find_label` scan of the current instruction list, so:

| goto shape | old interpreter |
|---|---|
| forward, same block | ✅ |
| backward (manual loop) | ❌ raised "goto target not found" |
| `continue` (out of an `if`) | ❌ raised |
| break-style (out of a loop) | ❌ raised |

So there was no correct oracle to mirror. This PR makes `goto`/`label` **correct on both engines**.

## Approach

A tail-recursive list walker has already consumed the block head at a goto, so labels are resolved **ahead of execution** rather than scanned at runtime. Both engines have analogous `cont` nesting (`:test` pushes one entry, loops two), so a goto resolves to `{depth-to-unwind, target, close-level}` on each — differing only in the target representation.

- **Interpreter** (`Lua.Compiler.GotoResolution`, runs after codegen): rewrites `{:goto, name}` → `{:goto, id}` and records `proto.goto_targets[id] = {depth, level, target_tail}`. The goto handler drops `depth` `cont` entries, closes upvalues `>= level`, and resumes `target_tail`.
- **Dispatcher** (`Bytecode.resolve_gotos/2`, over the encoded tuple tree): `:label` → `{@op_label, …}` no-op; `:goto` → `{@op_goto, depth, target_pc, level}`. The handler closes upvalues, unwinds `depth` `cont` entries (`unwind_goto/2`, reusing `:break`'s marker shape), and resumes the destination tuple at `target_pc`.
- **Close level**: scope resolution records each label's live-local register watermark on `{:label, name, level}`; a goto closes open upvalue cells `>= level`, matching Lua 5.3 §3.3.4 block-exit (verified via the upvalue-capture test).

`do` blocks are flat-inlined, so reused label names collapse into one list; resolution picks the nearest label after the goto (else before), matching lexical intent for the `continue`/`break`/backward-loop idioms.

## Coverage status (honest)

`goto`/`label` are now covered. **One gap remains that the original exploration missed**: short-circuit `and`/`or` (`:test_and` / `:test_or`) still falls back to the interpreter. It's now the single documented exception in the `fully_compiled?/1` corpus guard, left for a focused follow-up. So this PR closes the known gap but the dispatcher is not yet at literal 100%.

## Tests / verification

- New `test/lua/vm/goto_test.exs` — runs each program on **both** engines (bytecode stripped vs kept) and asserts equal, correct results: forward/backward/continue/break/nested/reused-label/upvalue-capture.
- `dispatcher_test.exs` — a `:goto / :label` block asserting `@op_goto` is encoded (helper now descends into nested body tuples) and runs correctly.
- `bytecode_test.exs` — goto programs added to the coverage corpus; fallback/cascade tests reworked around `and`/`or`.
- Full suite green: **2443 passed**, lua53 suite **17 passed**. Dispatcher correctness check: `fib(20) dispatcher == interpreter`. (Benchee not installed in this env, so the perf comparison wasn't run; the hot path gains only three jump-table clauses for a rare opcode.)

### goto.lua skip

goto.lua lines 163-195 (the `debug.upvalueid` cell-identity assertions) are now skipped: `debug.upvalueid` is a stub returning `nil`, so those assertions can't pass once the (now-correct) goto control flow reaches them. The control flow itself stays covered by the preceding `assert(#a == 6)`.

### DIFF
--- a/lib/lua/compiler.ex
+++ b/lib/lua/compiler.ex
@@ -8,6 +8,7 @@ defmodule Lua.Compiler do
   alias Lua.AST.Chunk
   alias Lua.Compiler.Bytecode
   alias Lua.Compiler.Codegen
+  alias Lua.Compiler.GotoResolution
   alias Lua.Compiler.Prototype
   alias Lua.Compiler.Scope
 
@@ -29,7 +30,16 @@ defmodule Lua.Compiler do
   def compile(%Chunk{} = chunk, opts \\ []) do
     with {:ok, scope_state} <- Scope.resolve(chunk, opts),
          {:ok, prototype} <- Codegen.generate(chunk, scope_state, opts) do
-      {:ok, Bytecode.compile(prototype)}
+      # Encode bytecode first (it reads the raw `:goto` / `:label` stream),
+      # then resolve gotos for the list interpreter. The two passes are
+      # independent: the dispatcher runs `bytecode`, the interpreter runs the
+      # resolved `instructions` plus `goto_targets`.
+      prototype =
+        prototype
+        |> Bytecode.compile()
+        |> GotoResolution.resolve()
+
+      {:ok, prototype}
     end
   end
 
--- a/lib/lua/compiler.ex
+++ b/lib/lua/compiler.ex
@@ -9,6 +9,7 @@ defmodule Lua.Compiler do
   alias Lua.Compiler.Bytecode
   alias Lua.Compiler.Codegen
   alias Lua.Compiler.GotoResolution
+  alias Lua.Compiler.GotoValidation
   alias Lua.Compiler.Prototype
   alias Lua.Compiler.Scope
 
@@ -28,7 +29,8 @@ defmodule Lua.Compiler do
   """
   @spec compile(Chunk.t(), compile_opts()) :: {:ok, Prototype.t()} | {:error, term()}
   def compile(%Chunk{} = chunk, opts \\ []) do
-    with {:ok, scope_state} <- Scope.resolve(chunk, opts),
+    with :ok <- GotoValidation.validate(chunk),
+         {:ok, scope_state} <- Scope.resolve(chunk, opts),
          {:ok, prototype} <- Codegen.generate(chunk, scope_state, opts) do
       # Encode bytecode first (it reads the raw `:goto` / `:label` stream),
       # then resolve gotos for the list interpreter. The two passes are
@@ -48,13 +50,13 @@ defmodule Lua.Compiler do
   """
   @spec compile!(Chunk.t(), compile_opts()) :: Prototype.t()
   def compile!(%Chunk{} = chunk, opts \\ []) do
-    # `compile/2`'s spec allows `{:error, _}` for forward compatibility, but the
-    # codegen path doesn't yet have an error-returning code path — codegen
-    # surfaces unsupported constructs by raising directly. Once codegen is
-    # converted to thread `{:ok, _} | {:error, _}` through every clause, swap
-    # this back to a `case` that re-raises `{:error, reason}` as a clear
-    # compiler exception.
-    {:ok, prototype} = compile(chunk, opts)
-    prototype
+    # Codegen surfaces unsupported constructs by raising directly, but the
+    # goto legality pass (`Lua.Compiler.GotoValidation`) returns `{:error,
+    # message}` for programs PUC-Lua rejects at compile time. Re-raise those as
+    # a clear compiler exception.
+    case compile(chunk, opts) do
+      {:ok, prototype} -> prototype
+      {:error, message} -> raise Lua.CompilerException, message
+    end
   end
 end
--- a/lib/lua/compiler/bytecode.ex
+++ b/lib/lua/compiler/bytecode.ex
@@ -68,10 +68,10 @@ defmodule Lua.Compiler.Bytecode do
   @op_numeric_for 36
 
   # B5c-v2: closures, upvalues, varargs, multi-return, loops, self, concat.
-  # The bitwise family and the `:set_list` multi-return tail (below) since
-  # joined the covered set, so the only opcodes the encoder still rejects
-  # are `:goto` / `:label` (label-resolution semantics not yet ported to the
-  # dispatcher's flat tuple model).
+  # The bitwise family, the `:set_list` multi-return tail (below), and
+  # `:goto` / `:label` (resolved to `@op_goto` / `@op_label` by
+  # `resolve_gotos/2`) have all joined the covered set — the encoder no
+  # longer rejects any opcode current codegen emits.
   @op_closure 37
   @op_set_upvalue 38
   @op_get_open_upvalue 39
@@ -104,6 +104,14 @@ defmodule Lua.Compiler.Bytecode do
   @op_bitwise_not 58
   @op_set_list_multi 59
 
+  # `:label` is a no-op that anchors a `goto` target PC. `:goto` is resolved at
+  # encode time to `{@op_goto, depth, target_pc, level}` (see `resolve_gotos/2`):
+  # `depth` is the `cont` entries the dispatcher drops to reach the target's
+  # block, `target_pc` the label's index there, `level` the upvalue-close
+  # threshold. Mirrors `Lua.Compiler.GotoResolution` for the interpreter.
+  @op_label 60
+  @op_goto 61
+
   @doc """
   Compile a prototype, populating its `bytecode` field on success.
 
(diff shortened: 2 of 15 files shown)


davydog187:
## 🤖 Automated multi-agent review — stabilization cycle

5 reviewers fanned out across correctness/conformance/dispatcher/tests/quality dimensions; **every finding was then adversarially verified** by an independent agent that ran repros against the branch tip and PUC-Lua. Raised **15** → **11 confirmed** (3 HIGH, 8 MEDIUM), **4 refuted**.

The 11 confirmed findings collapse into **4 distinct issues**. The first directly contradicts the PR's central "conformant goto/label" claim, so flagging it before merge.

---

### 🔴 HIGH — Flat-inlined `do` blocks break `goto` label visibility (§3.3.4), on **both** engines

Codegen flat-inlines `do … end` (`gen_statement` for `Statement.Do` appends `body ++ [close_upvalues]` with **no wrapper tuple**), so a `::label::` inside a `do` block becomes a plain `{:label, name, level}` in the *parent's* flat list. Both resolvers — `GotoResolution.find_label_pos/3` (interpreter) and `Bytecode.find_label/3` (dispatcher) — then apply a scope-blind *"nearest label after the goto, else nearest before"* heuristic. Per §3.3.4 a goto may only see labels in its **own or an enclosing** block, never a nested or sibling one. The moduledoc's claim that "forward-then-backward matches lexical intent" is the precise wrong assumption.

Three independently verified repros (run on branch tip, both engines, vs PUC-Lua):

```lua
-- 1) jump INTO a nested block's label
local x=0; goto l; do ::l:: x=100 end; ::l:: x=x+5; return x
-- PUC-Lua: 5   |  this PR: 105  (both engines)

-- 2) backward goto silently binds to a FORWARD sibling-block label
local out=0
do ::a:: out=out+1 if out<2 then goto a end end
do goto a out=out+100 ::a:: out=out+1000 end
return out
-- PUC-Lua: 1002  |  this PR: 1001  (both engines — block-1's loop never runs)

-- 3) sibling-block reuse
local a,b=0,0
do ::l:: a=a+1 if a<2 then goto l end end
do ::l:: b=b+1 if b<3 then goto l end end
return a,b
-- PUC-Lua: 2, 3  |  this PR: 1, 3  (both engines)
```

Silent wrong results (skipped loop iterations), no error. **Fix direction:** preserve `do`-block boundaries through codegen (a scope marker / per-label scope id) so the resolver restricts candidates to the goto's own/enclosing real scope before applying nearest-match. The positional heuristic cannot distinguish a sibling/nested block once flattened. `Bytecode.resolve_gotos` needs the same fix.

> Note: `test/lua/vm/goto_test.exs:108` ("reused label name") only exercises the all-forward variant, so it passes *by luck* — neither backward-vs-forward-sibling collision nor jump-into-nested-block is covered.

---

### 🟠 MEDIUM — No static `goto`/label legality validation (§3.3.4)

PUC-Lua rejects these at **compile time**; this impl silently accepts and runs them (both engines). Nothing in the parser, `Scope.resolve`, or `GotoResolution` validates legality (`scope.ex` only records the label's register watermark).

```lua
-- jump into the scope of a local — reads uninitialised/stale register
do goto skip; local y = 10; ::skip:: return y end   -- PUC: compile error | PR: [nil]
-- with register reuse this can surface an unrelated live value, not just nil

-- duplicate label in the same scope
::a:: ::a:: return 1                                  -- PUC: "label 'a' already defined" | PR: [1]
```

**Fix direction:** a label-legality pass (in `Scope.resolve` or dedicated) that rejects jump-into-local-scope and duplicate labels with PUC-shaped messages, running before both resolvers.

---

### 🟠 MEDIUM — Undefined/unreachable `goto` crashes the compiler with an internal error, and engines diverge

`Bytecode.resolve_gotos` → `resolve_entry/4` does `raise "unresolved goto target … — codegen should not emit this"`. But codegen **does** emit it: the parser accepts gotos to undefined/out-of-scope labels, so this is reachable from ordinary source.

```lua
goto nowhere; return 1
-- PUC: "no visible label 'nowhere' for <goto>" (clean compile error)
-- PR (bytecode path): RuntimeError "unresolved goto target \"nowhere\" — codegen should not emit this" at COMPILE time
-- PR (interpreter fallback, e.g. with an and/or present): "goto target not found" at RUN time
```

The two passes disagree on the same illegal input (compile-time bare raise vs. runtime `InternalError`), and `GotoResolution.resolve_instr`'s `:error` branch + comment ("parser should have rejected this") is effectively dead code on the default path. **Fix direction:** one shared legality pass returning a proper `{:error, CompileError}` with a PUC-shaped message for both engines; drop the false "codegen should not emit this" invariant.

---

### 🟠 MEDIUM — `goto.lua` suite skip is mis-categorized (masks the gap above)

`test/lua53_skips.exs` skips `goto.lua` lines 12–40 with reason *"load() parse-error messages do not match PUC format"* (category `:stdlib`). But those lines use `errmsg(...)` which **first requires `load()` to fail** — they test illegal-goto *rejection*, not message wording. Since this impl doesn't reject (repeated-label / jump-into-block / jump-over-local all load & run), the `:stdlib` "message format" reason hides a real §3.3.4 conformance gap from suite-status tracking. **Fix:** re-categorize as `:unimplemented` with an honest reason (+ tracking issue), or implement the checks.

---

<details><summary>4 findings considered and refuted (with reasoning)</summary>

- **"PR-claimed rejections have no test coverage"** — the PR body claims no such rejections; it scopes itself to *valid* goto shapes. The static-validation gap is real but pre-existing/out of scope (captured above as MEDIUMs).
- **"Corpus guard only asserts compilability"** — by design; execution correctness for every shape (incl. continue/break) is value-checked in `goto_test.exs` via `assert_both/2`.
- **"Executor comment documents a 2-tuple, code stores 3-tuple"** — real but cosmetic comment drift at `executor.ex:657`; the authoritative `GotoResolution` moduledoc is correct; zero functional impact.
- **"goto.lua skip masks untested upvalue-cell freshness"** — the close-level feature *is* value-tested on the forward-exit path; only the distinct-cells-across-backward-passes assertion is missing (minor coverage nicety).

</details>

---

**Bottom line:** the valid goto shapes the PR targets work and are well-tested, but the "conformant" framing is too strong — the do-block flattening (🔴) produces silently wrong results on legal programs, and static legality (🟠) is unenforced. Recommend addressing the HIGH (or narrowing the PR's conformance claim + tracking the rest) before merge.

I can implement any of these fixes if you'd like — say the word.

<sub>Generated by an automated multi-agent review+verify workflow. Findings were verified against branch tip and PUC-Lua but may still contain errors; treat as review signal, not ground truth.</sub>

davydog187:
## ✅ Conformance fixes landed (commits `82281ab`, `009b880`)

The HIGH + MEDIUM findings from the review above are fixed and **independently verified against PUC-Lua 5.5 on both engines** (not just self-reported green):

**Approach:** lexical **block-path** tracking — `Scope` threads a per-`do`-block id path; codegen emits `{:label, name, level, block_path}` / `{:goto, name, block_path}`; both resolvers (`GotoResolution` interpreter, `Bytecode.resolve_gotos` dispatcher) filter candidates by a `visible?/2` prefix test (a label is visible iff its block path is a prefix of the goto's — own/enclosing block, never nested or sibling). A new `Lua.Compiler.GotoValidation` AST pass runs first in `Compiler.compile` and rejects illegal gotos with PUC-shaped messages.

**Verified (interpreter == dispatcher == PUC):**

| repro | before | after / PUC |
|-------|--------|-------------|
| jump into nested `do` label | `105` | **`5`** |
| backward goto vs forward sibling label | `1001` | **`1002`** |
| sibling-block reuse | `[1,3]` | **`[2,3]`** |

**Rejections now compile-error (both engines), matching PUC wording:**
- `goto nowhere` → `no visible label 'nowhere' for goto`
- `::a:: ::a::` → `label 'a' already defined`
- `do goto skip; local y=10; ::skip:: end` → `jumps into the scope of local 'y'`

This also removes the bare `"codegen should not emit this"` internal raise (BUG 3) — undefined/out-of-scope gotos are now a clean compile error consistently across both engines — and the mis-categorized `goto.lua` suite skip (BUG 4): the illegal-goto rejection block (lines 12–40) now runs and passes; only the `debug.upvalueid` range stays skipped.

**Verification:** `mix test` 2450 passed · `mix test --only lua53` 17 passed (goto.lua's `errmsg` block now exercised) · `mix format` clean · `mix compile --warnings-as-errors` clean · **Dialyzer clean** · **full CI green**.

<sub>Fixes implemented by an automated agent, then independently verified against PUC-Lua (`lua-5.5`/`luac-5.5`) and CI by the orchestrator. The first push tripped CI Dialyzer (a dead `format_load_error` clause); `009b880` fixes it.</sub>

PR #366, opened 2026-06-15
fix(parser): propagate deep errors past function-call argument lists

davydog187:
## Problem

A parse error deep inside a function-call argument was reported at the **opening** line of the call, not where the mistake actually is. Given:

```lua
w:step("w.h42zel", "code", "My Code", { continueOnError = false }, function(output)
    local usercode = function()
      device.doesntExist)   -- line 3: the real error
    end
    return usercode()
end)
```

the parser emitted `Expected :delimiter::rparen, got :keyword::function` pinned at line 1 (the `w:step(` opener) instead of line 3.

## Root cause

In `parse_expr_list_acc/2`, when an argument sub-parse fails, `committed?/2` decides whether the failure is a real deep error to propagate or a clean signal that the argument list ended. Its fallback clause derives "committed" by comparing the error's byte offset against the start token via `error_position/1`, which only understands `:unexpected_token` errors:

```elixir
defp error_position({:unexpected_token, _type, pos, _message}), do: pos
defp error_position(_), do: nil
```

The actual failure is a `{:bare_expression, …}` error (`device.doesntExist` is a bare table access, not a valid statement). That shape returned `nil`, so the `committed?` fallback reported "not committed", the list "recovered" as if the arguments ended just before `function`, and `expect_closing` produced the misleading "Expected rparen" at the opener. This is exactly the failure mode the comment at `parse_expr_list_acc/2` warns about — it just never covered non-`:unexpected_token` shapes.

## Fix

Special-case the position-independent error shapes in `committed?/2` so they are *always* committed, rather than position-gating them through `error_position/1`:

```elixir
defp committed?({:unexpected_end, _message, _pos}, _tokens), do: true
defp committed?({:bare_expression, _pos, _expr_struct}, _tokens), do: true
defp committed?({:invalid_assign_target, _pos, _expr_struct}, _tokens), do: true
defp committed?({:unclosed_delimiter, _delimiter, _open_pos}, _tokens), do: true
```

These shapes can never represent a clean list terminator:

- `:bare_expression` / `:invalid_assign_target` are statement-level errors only reachable through a `function() … end` body — i.e. only after a full prefix expression has been consumed.
- `:unclosed_delimiter` has by construction consumed an opener and its content. Gating it on position would make blame depend on whether the opener happens to be the argument's first token versus opened mid-element; always committing it blames the inner opener uniformly.

Committing them unconditionally also sidesteps the missing-position problem: an invalid-assign target on a `meta: nil` postfix node (e.g. an `Expr.Call` LHS) has no `target_position/1`. To keep `convert_error/2` pointing at the right line, `validate_assign_targets/2` now takes the `=` operator position and uses it as a fallback:

```elixir
other -> {:halt, {:error, {:invalid_assign_target, target_position(other) || assign_pos, other.__struct__}}}
```

`committed?/2` now classifies these as committed and the genuine deep error propagates. The example above now reports:

```
at line 3, column 7:
A bare table-access expression isn't a Lua statement.
```

This is the inverse companion to #365 ("blame the opener"): don't let a shallow opener-blame swallow a genuine deep error.

## Tests

- New regression rows in the `@deep_propagation_errors` table in `test/lua/parser/error_position_test.exs`, covering each newly-committed shape (bare expression, invalid assign target with a position-less `Expr.Call` node, and unclosed delimiters both mid-element and beginning a non-first argument).
- Full suite green: `mix test` → 2475 passed, 7 skipped.

### DIFF
--- a/lib/lua/parser.ex
+++ b/lib/lua/parser.ex
@@ -1254,6 +1254,9 @@ defmodule Lua.Parser do
   end
 
   defp error_position({:unexpected_token, _type, pos, _message}), do: pos
+  defp error_position({:bare_expression, pos, _expr_struct}), do: pos
+  defp error_position({:invalid_assign_target, pos, _expr_struct}), do: pos
+  defp error_position({:unclosed_delimiter, _delimiter, open_pos}), do: open_pos
   defp error_position(_), do: nil
 
   defp parse_expr_list_until(tokens, terminator, delimiter, open_pos) do
--- a/lib/lua/parser.ex
+++ b/lib/lua/parser.ex
@@ -642,7 +642,7 @@ defmodule Lua.Parser do
   end
 
   defp parse_assignment(targets, [{:operator, :assign, pos} | rest]) do
-    case validate_assign_targets(targets) do
+    case validate_assign_targets(targets, pos) do
       :ok ->
         case parse_expr_list(rest) do
           {:ok, values, rest2} ->
@@ -669,13 +669,17 @@ defmodule Lua.Parser do
   # Name | prefixexp '[' exp ']' | prefixexp '.' Name. Anything else
   # (number, string, function call, parenthesised expression) is a
   # syntax error.
-  defp validate_assign_targets(targets) do
+  # `assign_pos` is the position of the `=` operator, used as a fallback when
+  # the offending target node carries no position (e.g. an `Expr.Call` LHS,
+  # built with `meta: nil`) so the error still points at the right line rather
+  # than degrading to a nil position.
+  defp validate_assign_targets(targets, assign_pos) do
     Enum.reduce_while(targets, :ok, fn target, :ok ->
       case target do
         %Expr.Var{} -> {:cont, :ok}
         %Expr.Property{} -> {:cont, :ok}
         %Expr.Index{} -> {:cont, :ok}
-        other -> {:halt, {:error, {:invalid_assign_target, target_position(other), other.__struct__}}}
+        other -> {:halt, {:error, {:invalid_assign_target, target_position(other) || assign_pos, other.__struct__}}}
       end
     end)
   end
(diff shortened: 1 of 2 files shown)


davydog187:
## Stabilization cycle (ultracode review)

Ran a multi-agent review (4 dimensions: regression-risk, edge-cases, completeness, test-coverage) with adversarial verification (each finding refuted by 2 independent skeptics) plus a validation gate. 11 findings, 7 confirmed. Results actioned in 5e73e77:

**Blocker — formatting.** The new test tuple wasn't `mix format`-clean. Fixed.

**High — nil-position fall-through (real bug, same class as the original).** The original `:invalid_assign_target` clause only helped when `target_position/1` returned a real position — but it returns `nil` for an `Expr.Call`/`Expr.Paren` LHS (those nodes are built with `meta: nil`). So this still reproduced the bug:

```lua
outer("ok", function()
  f() = 1        -- blamed the `function` keyword on line 1, not here
end)
```

Fix: classify `:bare_expression` and `:invalid_assign_target` as *always-committed* in `committed?/2` (they're statement-level errors only reachable through a function-literal body, so they always represent consumed input — like `:unexpected_end`), independent of position. Also threaded the `=` operator position into `validate_assign_targets/2` as a fallback so the propagated error gets a real line instead of `nil`. Now reports `syntax error near '='` at line 2.

**Medium — `:unclosed_delimiter` clause was untested.** Deleting it kept all tests green. Added a row for a non-first-argument unclosed delimiter (`outer(1, inner(`), plus a row for the invalid-assign case above.

**Not actioned (intentional):**
- *`{`/`(`-prefix unclosed argument inconsistency* (low): an unclosed delimiter that *begins* an argument is still blamed at the outer boundary because its opener position coincides with the sub-parse start token. Relaxing `>` to `>=` in `committed?/2` would change this, but it's a broader behavior change with its own trade-offs — left as a possible follow-up, not folded into this PR.

Validation after changes: `mix test` → 2477 passed, 7 skipped; `mix format --check-formatted` clean; credo not configured in this repo.

davydog187:
Folded the deferred follow-up into this PR (65a7c04).

**`{`/`(`-prefix unclosed argument inconsistency** — now addressed. Rather than the broad `>` → `>=` relaxation in `committed?/2` (which would also make `:unexpected_token` errors at the start token "committed" and break clean trailing-comma / stray-closer recovery, e.g. `f(1, )` and `f(1, ]`), `:unclosed_delimiter` is now made *unconditionally committed* — joining `:unexpected_end`, `:bare_expression`, and `:invalid_assign_target`. An opened-but-unclosed delimiter has by construction consumed an opener and content, so it can never be a clean list terminator and always propagates as the genuine deep error. `:unexpected_token` keeps its position-gated recovery.

Result — an unclosed delimiter beginning a non-first argument is now blamed at its own opener, consistent with the first-arg and mid-element cases:

| source | before | after |
|---|---|---|
| `outer(1, {a = 1` | `Expected rparen, got lbrace` | `Unclosed opening brace '{'` @ col 10 |
| `outer(1, (2 + 3` | `Expected rparen, got lparen` | `Unclosed opening parenthesis` @ col 10 |

Two regression rows added to `@deep_propagation_errors`. Validation: `mix test` → 2479 passed, 7 skipped; `mix format --check-formatted` clean.

PR #375, opened 2026-06-20
fix(vm): encode Elixir nil to Lua nil instead of the string "nil"

### LINKED ISSUE #374: `Lua.encode!/2` encodes Elixir `nil` as the Lua string "nil" instead of Lua `nil` (asymmetric with decode)
We're testing this against:

* Lua 1.0.0-rc.3

`encode!/2` encodes atoms as Lua strings and doesn't exempt nil, so Lua.encode!(lua, nil) yields the Lua string "nil" rather than Lua nil. 

```elixir
# 1. Top-level encode: produces a string, not Lua nil
Lua.encode!(Lua.new(), nil)
#=> {"nil", #Lua<>}      # expected: {nil, #Lua<>}

# 2. Round trip is lossy
{enc, lua} = Lua.encode!(Lua.new(), nil)
Lua.decode!(lua, enc)
#=> "nil"                # expected: nil

# 3. ...yet decode maps Lua nil to Elixir nil (the asymmetry)
Lua.eval!(Lua.new(), "return nil") |> elem(0)
#=> [nil]
```

What we'd expect is `Lua.encode!(lua, nil) == {nil, lua}`, so `decode(encode(nil)) == nil`, matching decode's treatment of Lua `nil`. What's happening is `encode!/2` returns `{"nil", lua}`; the round trip returns the string `"nil"`; and the encoded value is truthy in Lua.

The impact is that APIs that return (value, error) pairs — e.g. `return nil, "not found"` — come back to Lua scripts as the truthy string "nil", so `if not value then ... end` takes the wrong branch instead of handling the miss. It fails quietly (no error) and is data-dependent, which makes it hard to spot.

Here's a unit test that highlights the issue:

```elixir
defmodule Turn.Apps.LuaApp.LuaEncodeNilTest do
  @moduledoc """
  Standalone reproduction of an asymmetric `nil` encoding in the `lua` package
  (1.0.0-rc.3).

  Elixir's `nil` is the atom `:nil`. `Lua.encode!/2` encodes atoms as Lua
  strings, and does not exempt `nil`, so `Lua.encode!(lua, nil)` produces the Lua
  string `"nil"` rather than Lua `nil`. Two problems follow:

    * **Asymmetry** — `decode` maps Lua `nil` back to Elixir `nil`, so
      `decode(encode(nil))` is `"nil"`, not `nil`: the round trip is lossy for
      the one value a bridge most needs to preserve.

    * **Silent inversion** — every Lua string (including `"nil"`) is truthy, so a
      value that meant "absence" becomes truthy. Code like
      `local v = api(); if not v then ... end` takes the wrong branch instead of
      erroring.

  The behaviour is also internally inconsistent: the VM already maps `nil` to Lua
  `nil` when it appears *inside a table* or *in a function's result list* — only
  the top-level scalar `encode!/2` diverges (see the last two tests, which pass
  today).

  The first three tests are expected to pass once `encode!/2` maps `nil` to Lua
  `nil`; today they fail. The last two already pass and show the inconsistency.
  """
  use ExUnit.Case, async: true

  describe "the bug (these fail against 1.0.0-rc.3)" do
    test "encode!/2 maps Elixir nil to Lua nil, not the string \"nil\"" do
      # Today this returns {"nil", _lua}.
      assert {nil, _lua} = Lua.encode!(Lua.new(), nil)
    end

    test "decode(encode(nil)) round-trips to nil" do
      {encoded, lua} = Lua.encode!(Lua.new(), nil)
      # Today `encoded` is "nil", so this decodes to the string "nil".
      assert Lua.decode!(lua, encoded) == nil
    end

    test "an encoded nil is falsy in Lua" do
      {encoded, lua} = Lua.encode!(Lua.new(), nil)
      lua = Lua.set!(lua, [:value], encoded)

      {[branch], _lua} = Lua.eval!(lua, "if value then return 'truthy' else return 'falsy' end")

      # Today this is "truthy": the string "nil" is truthy in Lua, so "absence"
      # silently flips control flow.
      assert branch == "falsy"
    end
  end

  describe "the inconsistency (these already pass, showing nil↔nil is the intended model)" do
    test "decode maps Lua nil to Elixir nil" do
      assert {[nil], _lua} = Lua.eval!(Lua.new(), "return nil")
    end

    test "nil inside a table is already encoded as Lua nil" do
      {encoded, lua} = Lua.encode!(Lua.new(), %{"present" => 1, "absent" => nil})
      lua = Lua.set!(lua, [:t], encoded)

      assert {["number,nil"], _lua} =
               Lua.eval!(lua, "return type(t.present) .. ',' .. type(t.absent)")
    end

    test "a raw nil in a function's result list is already treated as Lua nil" do
      lua = Lua.set!(Lua.new(), [:f], fn _args, state -> {[nil], state} end)
      assert {["nil"], _lua} = Lua.eval!(lua, "return type(f())")
    end
  end
end
```

smn:
Fixes #374

full disclosure: Claude helped here

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- `Lua.encode!/2` now maps Elixir `nil` to Lua `nil` instead of the string
+  `"nil"`. Previously top-level `nil` fell into the atom-encoding head and
+  became `"nil"`, which is truthy in Lua — silently inverting `if not value
+  then ...` checks and breaking `return nil, "reason"` error patterns. The
+  round trip `decode!(encode!(nil))` is now lossless, matching the existing
+  behaviour for `nil` inside tables and function result lists (#374).
+
 ## [1.0.0-rc.3] - 2026-06-15
 
 The fourth release candidate for `1.0.0`. It builds on rc.2 with a
--- a/lib/lua.ex
+++ b/lib/lua.ex
@@ -917,6 +917,14 @@ defmodule Lua do
       true
   """
   @spec encode!(t(), term()) :: {term(), t()}
+  # Elixir `nil` maps to Lua `nil`, mirroring `decode!/2` so the round trip is
+  # lossless. Without this clause `nil` falls into the atom head below and
+  # encodes to the string `"nil"`, which is truthy in Lua and silently inverts
+  # `if not value then ...` checks (e.g. `return nil, "not found"` patterns).
+  def encode!(%__MODULE__{} = lua, nil) do
+    {nil, lua}
+  end
+
   def encode!(%__MODULE__{} = lua, value) when is_atom(value) and not is_boolean(value) do
     {Atom.to_string(value), lua}
   end
--- a/test/lua_test.exs
+++ b/test/lua_test.exs
@@ -722,6 +722,23 @@ defmodule LuaTest do
       assert [{1, 1}, {2, 2}] = lua |> Lua.decode!(ref) |> Enum.sort()
     end
 
+    test "it encodes nil to Lua nil rather than the string \"nil\"" do
+      lua = Lua.new()
+
+      # Top-level nil must map to Lua nil, not the truthy string "nil"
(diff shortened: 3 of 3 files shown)


davydog187 (approved):
Amazing!

PR #376, opened 2026-06-20
fix(string.rep): allocate proportionally to result, not repeat count

### LINKED ISSUE #373: `Lua.call_function!/3` allocates ~50× the payload when a Lua function returns a large string
First of all, **thanks** for all the work on this package :) We're currently running `0.4.0` with luerl and are taking 1.0.0-rc3 through it's paces.

We're testing this on:

* Lua 1.0.0-rc.3

What we're seeing is that producing a large string with `Lua.eval!/2` allocates memory proportional to the string. Returning the same string from a Lua function invoked via `Lua.call_function!/3` allocates roughly 50–60× as much — a 16 MB string allocates ~1 GB. On a process with `max_heap_size` set (a common pattern for sandboxing untrusted scripts), this trips the limit and kills the process.

to reproduce use the following:

```elixir
size = 16 * 1024 * 1024

measure = fn fun ->
  :erlang.garbage_collect()
  before = :erlang.memory(:total)
  fun.()
  div(:erlang.memory(:total) - before, 1024 * 1024)
end

# Proportional: ~16 MB
measure.(fn -> Lua.eval!(Lua.new(), ~s[return string.rep("x", #{size})]) end)

# Excessive: ~1006 MB for the same 16 MB string
{[fun], lua} = Lua.eval!(Lua.new(), ~s[return function() return string.rep("x", #{size}) end])
measure.(fn -> Lua.call_function!(lua, fun, []) end)
```

We would expect that `call_function!/3` returning a string should allocate proportionally to that string, like `eval!/2` does. What actually happens is that the 16 MB return allocates ~1 GB. We observed `eval!/2` to use ~16 MB while `call_function!/3` used ~1006 MB. The amplication seems to get worse as more is loaded into the VM with `load_api/2`.

Here's a test to reproduce:

```elixir
defmodule Turn.Apps.LuaApp.LuaCallFunctionMemoryTest do
  @moduledoc """
  Standalone reproduction of a memory blowup in the `lua` package (1.0.0-rc.3)
  when a large string is returned through `Lua.call_function!/3`.

  Producing a large string with `Lua.eval!/2` allocates an amount of memory
  proportional to the string. Returning that *same* string from a Lua function
  invoked via `Lua.call_function!/3` allocates ~50–60x as much (a 16 MB string
  allocates ~1 GB), which on a heap-limited process trips `max_heap_size` and
  kills it.
  """
  use ExUnit.Case, async: false

  # 16 MB — the smallest size that reliably shows the blowup.
  @size 16 * 1024 * 1024

  # Allocating up to this many bytes is considered "proportional". The blowup
  # allocates ~1 GB, far above this ceiling.
  @reasonable_ceiling 200 * 1024 * 1024

  defp peak_allocation(fun) do
    :erlang.garbage_collect()
    before = :erlang.memory(:total)
    result = fun.()
    delta = :erlang.memory(:total) - before
    {result, delta}
  end

  test "Lua.eval!/2 returning a large string allocates proportionally" do
    {{[string], _lua}, delta} =
      peak_allocation(fn ->
        Lua.eval!(Lua.new(), ~s[return string.rep("x", #{@size})])
      end)

    assert byte_size(string) == @size
    assert delta < @reasonable_ceiling
  end

  # This is the failing case: the same string returned through a function call
  # allocates an order of magnitude more memory than `Lua.eval!/2` above.
  test "Lua.call_function!/3 returning a large string allocates excessively" do
    {[fun], lua} = Lua.eval!(Lua.new(), ~s[return function() return string.rep("x", #{@size}) end])

    {{[string], _lua}, delta} =
      peak_allocation(fn ->
        Lua.call_function!(lua, fun, [])
      end)

    assert byte_size(string) == @size

    # Expectation for a fixed library: this allocation should be proportional to
    # the returned string, like `Lua.eval!/2`. Today it allocates ~1 GB.
    assert delta < @reasonable_ceiling,
           "Lua.call_function!/3 allocated #{div(delta, 1024 * 1024)} MB to return a " <>
             "#{div(@size, 1024 * 1024)} MB string"
  end
end
```

smn:
Fixes #373

full disclosure: Claude helped here

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,9 @@ jobs:
       matrix:
         include:
           - elixir: "1.19.0"
-            otp: "28.0"
+            # 28.1+, not 28.0: Hex 2.5.0 (setup-beam's autoinstalled default)
+            # calls :re.import/1, which only exists in OTP 28.1+.
+            otp: "28.1"
             bootstrap_beam: false
           # OTP 29 RC's :httpc can't reach builds.hex.pm, so we tell setup-beam
           # to skip Hex/Rebar autoinstall and curl them ourselves below.
@@ -92,7 +94,9 @@ jobs:
       matrix:
         include:
           - elixir: "1.19.0"
-            otp: "28.0"
+            # 28.1+, not 28.0: Hex 2.5.0 (setup-beam's autoinstalled default)
+            # calls :re.import/1, which only exists in OTP 28.1+.
+            otp: "28.1"
             bootstrap_beam: false
 
     steps:
@@ -126,7 +130,9 @@ jobs:
       matrix:
         include:
           - elixir: "1.19.0"
-            otp: "28.0"
+            # 28.1+, not 28.0: Hex 2.5.0 (setup-beam's autoinstalled default)
+            # calls :re.import/1, which only exists in OTP 28.1+.
+            otp: "28.1"
             bootstrap_beam: false
 
     steps:
@@ -224,7 +230,9 @@ jobs:
       matrix:
         include:
           - elixir: "1.19.0"
-            otp: "28.0"
+            # 28.1+, not 28.0: Hex 2.5.0 (setup-beam's autoinstalled default)
+            # calls :re.import/1, which only exists in OTP 28.1+.
+            otp: "28.1"
             bootstrap_beam: false
 
     steps:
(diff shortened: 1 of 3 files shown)


davydog187:
Looks great @smn. Will merge once CI passes

PR #379, opened 2026-07-03
fix(lua): hand callbacks Lua.t regardless of entry point

### LINKED ISSUE #377: Two-arity callbacks receive raw `Lua.VM.State` (not `Lua.t`) when entered via `Lua.encode!/2` or nested in a `set!` value
Again, thanks for the work on this package :)

We're testing this against:

* Lua 1.0.0-rc.3

I'm not sure if this is the intended behaviour but a two-arity Elixir callback (`fn args, state -> {results, state} end`) receives a **different** `state` argument — and must return a **different** state shape — depending on how it entered the VM:

| How the callback enters the VM                                | `state` it receives |
| ------------------------------------------------------------- | ------------------- |
| `Lua.set!(lua, [:f], fun)` — function directly at the path    | `Lua.t`             |
| `deflua` + `Lua.load_api/3`                                   | `Lua.t`             |
| `Lua.set!(lua, [:t], %{"f" => fun})` — function inside a value | raw `Lua.VM.State`  |
| `Lua.encode!(lua, fun)` — closure handed to Lua at runtime    | raw `Lua.VM.State`  |

The `Lua.set!/3` docs teach the `Lua.t` convention (treat `state` as a `Lua.t`, use the public API on it, return it as received). That holds for a function set directly at a path and for `deflua`. But a closure that reaches the VM as an **encoded value** — either via `Lua.encode!/2`, or nested inside a value passed to `set!` — is invoked with the raw `Lua.VM.State` instead.

The raw `Lua.VM.State` is unusable with the public API: `Lua.decode!/2`, `Lua.encode!/2`, `Lua.get_private!/2`, etc. all expect `Lua.t` and raise `no function clause matching`. And the two return conventions reject each other:

* a raw-state callback returning `{results, %Lua{}}` fails with `native function returned invalid result ..., expected {results, state}`
* a `set!`-at-path callback returning `{results, %Lua.VM.State{}}` fails with `deflua functions must return encoded data`

So the **same closure cannot work in both positions**. This bites any factory-style API — an Elixir function that builds a closure and hands it to Lua via `Lua.encode!/2` (e.g. a translator bound to a locale, or a store handle bound to a config). Written against the documented `set!` convention it works at a path and crashes as an encoded value:

```
** (Lua.RuntimeException) Lua runtime error: no function clause matching in Lua.get_private!/2
```

We'd expect a two-arity callback to receive `Lua.t` (and accept `{results, %Lua{}}` back) regardless of how it entered the VM, matching the documented `Lua.set!/3` convention. The workaround today is to hand-wrap the raw state as `%Lua{state: raw}` to reach the public API — which reaches into the struct's internals.

Here's a standalone test that pins the documented convention where it already holds (first block, passing) and asserts the same convention for the two encoded-value entry points (second block, failing as of 1.0.0-rc.3):

```elixir
defmodule Lua.CallbackStateAsymmetryTest do
  @moduledoc """
  A two-arity Elixir callback (`fn args, state -> {results, state} end`)
  receives a different `state` argument — and must return a different state
  shape — depending on how it entered the VM:

  | How the callback enters the VM                                  | `state` it receives |
  | --------------------------------------------------------------- | ------------------- |
  | `Lua.set!(lua, [:f], fun)` — function directly at the path       | `t:Lua.t/0`         |
  | `deflua` + `Lua.load_api/3`                                      | `t:Lua.t/0`         |
  | `Lua.set!(lua, [:t], %{"f" => fun})` — function inside a value   | raw `Lua.VM.State`  |
  | `Lua.encode!(lua, fun)` — closure handed to Lua at runtime       | raw `Lua.VM.State`  |

  The raw `Lua.VM.State` is unusable with the public API — `Lua.decode!/2`,
  `Lua.encode!/2`, `Lua.get_private/2`, etc. all expect `t:Lua.t/0` and raise
  "no function clause matching" — and the two return conventions reject each
  other:

    * a raw-state callback returning `{results, %Lua{}}` fails with
      `"native function returned invalid result ..., expected {results, state}"`
    * a `set!`-at-path callback returning `{results, %Lua.VM.State{}}` fails
      with `"deflua functions must return encoded data"`

  So the same closure cannot work in both positions. The `Lua.set!/3` docs
  teach the `t:Lua.t/0` convention, but any factory-style API — an Elixir
  function that builds a closure and returns it to Lua via `Lua.encode!/2`
  (e.g. a translator bound to a locale, a store handle bound to a config) —
  gets the raw state and must hand-wrap it in `%Lua{state: raw}` to use the
  public API, reaching into the struct's internals.

  The first describe block pins the documented `t:Lua.t/0` convention where it
  already holds (passing). The second asserts the same convention for the
  other two entry points and fails as of 1.0.0-rc.3.
  """
  use ExUnit.Case, async: true

  defmodule DefluaProbe do
    @moduledoc false
    use Lua.API, scope: "probe"

    deflua whoami(), state do
      {[inspect(state.__struct__)], state}
    end
  end

  # Reports the struct name of the state the VM handed the callback.
  defp probe_callback do
    fn _args, state -> {[inspect(state.__struct__)], state} end
  end

  # A callback written exactly the way the `Lua.set!/3` docs teach: treat
  # `state` as a `t:Lua.t/0`, use the public API on it, return it as received.
  defp documented_convention_callback do
    fn _args, state -> {[Lua.get_private!(state, :secret)], state} end
  end

  describe "callbacks that receive the public Lua.t (documented convention)" do
    test "a function set! directly at a path receives Lua.t" do
      lua = Lua.set!(Lua.new(), [:probe], probe_callback())

      assert {["Lua"], _} = Lua.eval!(lua, "return probe()")
    end

    test "a deflua function loaded via load_api receives Lua.t" do
      lua = Lua.load_api(Lua.new(), DefluaProbe)

      assert {["Lua"], _} = Lua.eval!(lua, "return probe.whoami()")
    end

    test "a set!-at-path callback can use the public Lua API on its state" do
      lua =
        Lua.new()
        |> Lua.put_private(:secret, "from-private")
        |> Lua.set!([:fetch], documented_convention_callback())

      assert {["from-private"], _} = Lua.eval!(lua, "return fetch()")
    end
  end

  describe "callbacks that enter the VM as encoded values (asymmetric)" do
    # FAILS: returns ["Lua.VM.State"]. Lua.encode!/2 stores the closure bare,
    # so the VM invokes it with its internal state rather than wrapping it the
    # way Lua.set!/3 does for a function set directly at a path.
    test "a closure embedded via encode! receives the same Lua.t" do
      {fun, lua} = Lua.encode!(Lua.new(), probe_callback())
      lua = Lua.set!(lua, [:probe], fun)

      assert {["Lua"], _} = Lua.eval!(lua, "return probe()")
    end

    # FAILS: returns ["Lua.VM.State"]. set! only wraps a function that is
    # itself the value at the path — a function reachable *inside* the value
    # (here, a table field) is encoded bare, like Lua.encode!/2 does.
    test "a function nested inside a table passed to set! receives the same Lua.t" do
      lua = Lua.set!(Lua.new(), [:api], %{"probe" => probe_callback()})

      assert {["Lua"], _} = Lua.eval!(lua, "return api.probe()")
    end

    # FAILS: the practical consequence. The identical callback — written
    # against the documented set! convention — works when registered at a path
    # and crashes when handed to Lua as an encoded value:
    #
    #   ** (Lua.RuntimeException) Lua runtime error:
    #      no function clause matching in Lua.get_private!/2
    #
    # And side-stepping the public API doesn't rescue it: returning the %Lua{}
    # shape from the encoded position trips "native function returned invalid
    # result", while returning the raw shape from the set! position trips
    # "deflua functions must return encoded data". No single closure satisfies
    # both entry points.
    test "the same callback works identically via set! and via encode!" do
      lua = Lua.put_private(Lua.new(), :secret, "from-private")

      lua = Lua.set!(lua, [:registered], documented_convention_callback())
      {encoded, lua} = Lua.encode!(lua, documented_convention_callback())
      lua = Lua.set!(lua, [:embedded], encoded)

      assert {["from-private", "from-private"], _} =
               Lua.eval!(lua, "return registered(), embedded()")
    end
  end
end
```

davydog187:
Fixes #377.

## Problem

A two-arity Elixir callback (`fn args, state -> {results, state} end`) received a **different** `state` — and had to return a different shape — depending on how it entered the VM:

| Entry point | `state` handed to the callback |
| --- | --- |
| `Lua.set!(lua, [:f], fun)` — fn at the path | `Lua.t` (`%Lua{}`) ✅ |
| `deflua` + `Lua.load_api/3` | `Lua.t` (`%Lua{}`) ✅ |
| `Lua.set!(lua, [:t], %{"f" => fun})` — fn nested in a value | raw `Lua.VM.State` ❌ |
| `Lua.encode!(lua, fun)` — closure encoded at runtime | raw `Lua.VM.State` ❌ |

`Lua.set!/3`'s `is_function` clauses wrap the raw `%Lua.VM.State{}` into the public `%Lua{}`, call the callback, unwrap the returned `%Lua{}` back to its `.state`, and validate the return is encoded. But functions reaching the VM as an **encoded value** — via `Lua.encode!/2` or nested inside a value passed to `Lua.set!/3` — go through `Lua.VM.Value.encode/2`, which stored them bare (`{:native_func, fun}`) with no wrapping. The VM then invoked them with the raw internal state, which the public API (`Lua.decode!/2`, `Lua.get_private!/2`, …) rejects with `no function clause matching`. So the same closure worked at a path and crashed as an encoded value.

## Fix

`Lua.VM.Value` is deliberately ignorant of `%Lua{}`, so the wrapping stays in the `Lua` layer and is now shared by every entry point:

- **`lib/lua.ex`** — extract the wrap/unwrap/validate closure builder into a shared `wrap_callback/3` (+ `validate_encoded!/3`). Both `set!`-at-path clauses use it, and it is injected into the two `Value.encode` call sites (`encode!/2` and the nested-value `set!` clause).
- **`lib/lua/vm/value.ex`** — `encode/3` takes an optional `fun_wrapper` threaded through the map/list recursion. The default preserves the prior raw behavior for low-level callers; the public `Lua` entry points supply the `%Lua{}` wrapper.

### Behavior changes (clean break, intended)

- 2-arity callbacks entered via `encode!/2` or nested in a `set!` value now receive `%Lua{}` and return `{results, %Lua{}}` (or bare results), matching the documented convention. Code relying on the rc.3 raw-state behavior or the `%Lua{state: raw}` workaround should drop the workaround.
- arity-1 callbacks entered via those paths now get the same return-encoding validation as the path entry point.

## Verification

- Added `test/lua/callback_state_asymmetry_test.exs` (the issue's test module). Its 3 encoded-value cases failed before the fix and pass after.
- `mix test --include lua53` — 2592 passed, `mix format --check-formatted`, `mix dialyzer` — 0 errors.

### DIFF
--- a/lib/lua.ex
+++ b/lib/lua.ex
@@ -303,67 +303,11 @@ defmodule Lua do
     raise Lua.RuntimeException, "Lua.set!/3 cannot have empty keys"
   end
 
-  def set!(%__MODULE__{} = lua, keys, func) when is_function(func, 1) do
+  def set!(%__MODULE__{} = lua, keys, func) when is_function(func, 1) or is_function(func, 2) do
     keys = keys |> List.wrap() |> Enum.map(&to_lua_key/1)
+    {function_name, scope} = List.pop_at(keys, -1)
 
-    wrapped =
-      {:native_func,
-       fn args, state ->
-         return = List.wrap(func.(args))
-
-         if not Util.list_encoded?(return) do
-           {function_name, scope} = List.pop_at(keys, -1)
-
-           raise Lua.RuntimeException,
-             function: function_name,
-             scope: scope,
-             message: "deflua functions must return encoded data, got #{inspect(return)}"
-         end
-
-         {return, state}
-       end}
-
-    state = do_set_nested(lua.state, keys, wrapped)
-    %{lua | state: state}
-  end
-
-  def set!(%__MODULE__{} = lua, keys, func) when is_function(func, 2) do
-    keys = keys |> List.wrap() |> Enum.map(&to_lua_key/1)
-
-    wrapped =
-      {:native_func,
-       fn args, state ->
-         {function_name, scope} = List.pop_at(keys, -1)
-
-         case func.(args, wrap(state)) do
-           {:error, reason, %__MODULE__{} = returned_lua} ->
-             raise RuntimeError, value: reason, state: returned_lua.state
-
-           {value, %__MODULE__{} = lua} ->
-             value = List.wrap(value)
-
-             if not Util.list_encoded?(value) do
-               raise Lua.RuntimeException,
-                 function: function_name,
-                 scope: scope,
-                 message: "deflua functions must return encoded data, got #{inspect(value)}"
-             end
-
-             {value, lua.state}
-
-           value ->
-             value = List.wrap(value)
-
-             if not Util.list_encoded?(value) do
-               raise Lua.RuntimeException,
-                 function: function_name,
-                 scope: scope,
-                 message: "deflua functions must return encoded data, got #{inspect(value)}"
-             end
-
-             {value, state}
-         end
-       end}
+    wrapped = wrap_callback(func, function_name, scope)
 
     state = do_set_nested(lua.state, keys, wrapped)
     %{lua | state: state}
@@ -372,12 +316,13 @@ defmodule Lua do
   def set!(%__MODULE__{} = lua, keys, value) do
     keys = keys |> List.wrap() |> Enum.map(&to_lua_key/1)
     value = Display.unwrap(value)
+    {function_name, scope} = List.pop_at(keys, -1)
 
     {encoded, state} =
       if Util.encoded?(value) do
         {value, lua.state}
       else
-        Value.encode(value, lua.state)
+        Value.encode(value, lua.state, &wrap_callback(&1, function_name, scope))
       end
 
     state = do_set_nested(state, keys, encoded)
@@ -935,7 +880,7 @@ defmodule Lua do
     if Util.encoded?(value) do
       {value, lua}
     else
-      {encoded, state} = Value.encode(value, state)
+      {encoded, state} = Value.encode(value, state, &wrap_callback(&1, :anonymous, []))
       {encoded, %{lua | state: state}}
     end
   rescue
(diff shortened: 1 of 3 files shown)


smn:
oh wow, awesome, thank you!

PR #383, opened 2026-07-08
fix(api): name the callee in call_function/3 errors

### LINKED ISSUE #92: Better error for undefined function
I often get `{:undefined_function, nil}` but it would be really helpful if we could include the function name which lua was trying to call

davydog187:
## Problem

Reported in #92. The rich in-Lua error path is already excellent (it names globals/locals/upvalues/fields/methods, renders source context, stack traces, and suggestions). The gap was entirely in the **programmatic** `Lua.call_function/3` boundary, which resolved a name itself and hand-rolled an error string that inspected the *resolved* value instead of the name that was looked up:

```elixir
{:error, "undefined function '#{inspect(func)}'", lua.state}  # reports nil, not the name
```

So `Lua.call_function(lua, [:bar], [])` returned `undefined function 'nil'`.

## Fix

Both fallbacks in `lib/lua.ex` now build the same `TypeError` the VM's `:call` opcode raises, via a shared `Executor.call_type_error/3`. Returned as a struct (not raised), so it flows through both existing contracts unchanged:

| Call | Before | After |
|------|--------|-------|
| `call_function(lua, [:bar], [])` | `{:error, "undefined function 'nil'", …}` | `{:error, "attempt to call a nil value (global 'bar')", …}` |
| `call_function(lua, [:x], [])` (x=5) | `{:error, "undefined function '5'", …}` | `{:error, "attempt to call a number value (global 'x')", …}` |
| `call_function!(lua, [:bar], [])` | `Lua runtime error: undefined function 'nil'` | Full render: message + `Suggestion:` block |
| nested `[:t, :missing]` | `undefined function 'nil'` | `attempt to call a nil value (field 'missing')` |

The terse (`pcall`-parity) and rich (`call_function!`) variants now match the in-Lua path exactly.

## Tests

- Resurrected `loading a file that calls an undefined function` as a real `load_file!` integration test (the `undefined_function.lua` fixture already existed).
- Covered the programmatic undefined/non-function path with new tests in `call_function_error_value_test.exs`.
- Deleted 6 redundant/invalid Luerl-era commented-out stubs whose paths `error_messages_test.exs` already covers, or whose premises are wrong for Lua 5.3.

All 2579 tests pass; `mix format` clean.

Closes #92

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -2,6 +2,9 @@ defmodule Lua.MixProject do
   use Mix.Project
 
   alias Lua.Parser.Error
+  alias Lua.VM.ArgumentError
+  alias Lua.VM.RuntimeError
+  alias Lua.VM.TypeError
   alias Mix.Tasks.Lua.Eval
 
   @url "https://github.com/tv-labs/lua"
@@ -17,6 +20,9 @@ defmodule Lua.MixProject do
     Lua.Chunk,
     Lua.RuntimeException,
     Lua.CompilerException,
+    RuntimeError,
+    TypeError,
+    ArgumentError,
     Error,
     Eval
   ]
@@ -53,18 +59,24 @@ defmodule Lua.MixProject do
         source_ref: "v#{@version}",
         # Render only the curated public surface; keep internals in source/IEx.
         filter_modules: fn module, _meta -> module in @public_modules end,
-        # `call_function/3` names the concrete VM exception structs it can hand
-        # back, but those modules are filtered above. Render the names as plain
-        # code instead of autolinking to (filtered) pages, which errors under
-        # `--warnings-as-errors`.
+        # The public VM exception structs' moduledocs name internal plumbing
+        # (`Lua.VM.Executor`, `Lua.VM.ErrorFormatter`) that stays filtered.
+        # Render those references as plain code instead of autolinking to
+        # filtered pages, which errors under `--warnings-as-errors`.
         skip_code_autolink_to: [
-          "Lua.VM.RuntimeError",
-          "Lua.VM.TypeError",
-          "Lua.VM.ArgumentError"
+          "Lua.VM.Executor.current_position/0",
+          "Lua.VM.ErrorFormatter.to_map/3"
         ],
         groups_for_modules: [
           Core: [Lua, Lua.API, Lua.Table, Lua.Chunk],
-          Errors: [Lua.RuntimeException, Lua.CompilerException, Error],
+          Errors: [
+            Lua.RuntimeException,
+            Lua.CompilerException,
+            RuntimeError,
+            TypeError,
+            ArgumentError,
+            Error
+          ],
           "Mix Tasks": [Eval]
         ],
         extras: [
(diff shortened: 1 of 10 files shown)


test/lua/call_function_error_value_test.exs
```diff
@@ -70,6 +72,63 @@ defmodule Lua.CallFunctionErrorValueTest do
...
 
+  describe "call_function/3 on a name that does not resolve to a function" do
+    test "an undefined name names what was looked up, not the resolved nil" do
+      {_, lua} = Lua.eval!(Lua.new(), "function foo() return 1 end")
+
+      assert {:error, reason, %Lua{}} = Lua.call_function(lua, [:bar], [])
+
+      # Regression: previously reported the resolved value ("undefined
+      # function 'nil'"). It must name the requested global instead.
+      assert reason == "attempt to call a nil value (global 'bar')"
```
davydog187: We should be returning the exception struct uniformly instead of returning a string. it should be up to the user to call `Exception.message()`. We should fix this across all of the public facing APIs so it consistently returns the excception struct

davydog187: Done in d6164a5. Both public `{:error, _}` APIs now hand back the exception struct uniformly:

- **`call_function/3`** → `{:error, exception, lua}` where `exception` is the VM struct (`Lua.VM.RuntimeError`/`TypeError`/`ArgumentError`). Per our discussion I wrapped *everything*, including user-thrown `error()` values — `error(42)` comes back as `%RuntimeError{value: 42}`, tables/`nil`/`false` likewise on `:value` — so pcall-parity data survives on the struct. In-Lua `pcall`/`xpcall` still project the raw §6.1 value via `ProtectedCall.error_value/1`; only the programmatic boundary changed.
- **`parse_chunk/1`** → `{:error, %Lua.CompilerException{}}` (was `{:error, [String.t()]}`); diagnostics on `:errors`.

Callers now own rendering via `Exception.message/1`. Doctests, the chunks guide, this regression suite, and a breaking-before-1.0 CHANGELOG entry are updated. Full suite (2577) + format + dialyzer green.

Those are the only two public APIs that returned error tuples — everything else (`eval!`, `load_file!`, `call_function!`) already raises, so callers get the struct there too.

mix.exs
```diff
@@ -53,6 +53,15 @@ defmodule Lua.MixProject do
...
         filter_modules: fn module, _meta -> module in @public_modules end,
+        # `call_function/3` names the concrete VM exception structs it can hand
+        # back, but those modules are filtered above. Render the names as plain
+        # code instead of autolinking to (filtered) pages, which errors under
+        # `--warnings-as-errors`.
+        skip_code_autolink_to: [
+          "Lua.VM.RuntimeError",
+          "Lua.VM.TypeError",
+          "Lua.VM.ArgumentError"
+        ],
```
davydog187: These need to be user facing exceptions. we don't want to do this.

davydog187: Fixed in b5c3057. Made `Lua.VM.RuntimeError`, `Lua.VM.TypeError`, and `Lua.VM.ArgumentError` user-facing: added them to `@public_modules` and the **Errors** doc group so they render on HexDocs (they're what `call_function/3` hands back).

That removed the need to `skip_code_autolink_to` *them* — those entries are gone. The only `skip_code_autolink_to` left is for the genuinely internal plumbing their moduledocs name (`Lua.VM.Executor.current_position/0`, `Lua.VM.ErrorFormatter.to_map/3`), which stays filtered.

`mix docs --warnings-as-errors` is green; full suite (2577) + format pass.

Issue #384, opened 2026-07-13
`Exception.message/1` on exception returns ANSI codes


kevinschweikert:
After #383 i tried it again in our codebase. Claude created this summary of the issue. I verified the reproduction.

VM error structs and `CompilerException` bake their fully-formatted (andANSI-colored) text into `:message` at construction time. Since that happens during VM execution — where `IO.ANSI.enabled?/0` is true — the escape codes and multi-line render survive into non-TTY sinks: log files, container`stdout`, Sentry/AppSignal titles. The formatter's TTY gate never fires at the point output is actually written.

#### Reproduction

```elixir
iex(1)> ex = Lua.VM.RuntimeError.exception(value: "kaboom", source: "demo", line: 1)
#Lua.VM.RuntimeError<
  value: "kaboom",
  lua_value: nil,
  source: "demo",
  message: "\e[2mat demo:1:\e[0m\n\n  runtime error: kaboom",
  call_stack: [],
  line: 1,
  ...
>
iex(2)> Application.put_env(:elixir, :ansi_enabled, false)
iex(3)> Exception.message(ex)
"\e[2mat demo:1:\e[0m\n\n  runtime error: kaboom"
```

#### Root cause

`ErrorFormatter.format/3` is correctly gated (`color/2` → `IO.ANSI.enabled?/0`,`error_formatter.ex:115`), but it's called eagerly and the result is memoized:

- `runtime_error.ex:42` — `message = ... || format_message(...)` (same in `assertion_error.ex`, `type_error.ex`, `argument_error.ex`, `internal_error.ex`)
- `compiler_exception.ex:16-18` — `exception/1` stores `Util.format_error/1` output as strings in :errors, so `parse_chunk/1`'s exception carries pre-rendered text too.

Freezing a TTY-dependent render at build time defeats the gate.

### Expectations

No newlines and ANSI codes in an exception message, to cleanly  log the errors in our log system

kevinschweikert:
For a nice log i currently have to do:

```elixir
Exception.message(ex)
|> String.replace(~r/\e\[[0-9;]*m/, "")
|> String.split("\n")
|> Enum.map(&String.trim/1)
|> Enum.reject(&(&1 == ""))
|> Enum.join()
```

PR #387, opened 2026-07-15
docs: fix exception-handling examples and Lua.new option doctests

davydog187:
## What

Fixes three classes of documentation mistake around exception handling and doctest rendering. All examples were verified by running them.

### 1. README: `e.message` returns `nil`

The "Error messages with source and line" tour rescued a `Lua.RuntimeException` from `error("...")` and read `e.message`. Since #384, that field is `nil` when the exception wraps a VM error — the message is rendered lazily via `Exception.message/1`. Switched the example to the accessor.

The illustrative comment was also wrong: it showed a `Runtime Error` title line the formatter never emits. Corrected to the actual (ANSI-elided) frame:

```
Lua runtime error: at <eval>:2:

  runtime error: something went wrong
```

`e.line`/`e.source` were already correct and are unchanged.

### 2. `Lua.new/1` option doctests rendered as plain text

The three `## Options` doctests (`:max_call_depth`, `:max_string_bytes`, `:max_instructions`) were indented 6 spaces. Nested under 2-space bullets, a code block needs **8 spaces** — so Markdown/ExDoc rendered them as lazy continuation paragraphs instead of highlighted code blocks. Re-indented all three (confirmed via `mix docs` that each now renders inside a `<pre>` block within its bullet).

Also replaced the double-escaped `\"x\"` in the `:max_string_bytes` example with a single-quoted Lua string (`'x'`) so it renders cleanly.

### 3. Marketing snippets used a function that doesn't exist

`website/.../snippets.ex` called `Lua.eval/2` (only `Lua.eval!/2` exists) in three places, and the sandbox snippet read the nil `.message` field with the wrong expected substring. Rewrote to idiomatic `Lua.eval!/2` (correct `{results, lua}` shape; the sandbox one now shows the raise, matching `os.execute(_) is sandboxed`).

## Verification

- `mix test --only doctest` — 49 pass (the three re-indented `Lua.new/1` doctests included).
- `mix docs` — the three option examples now render as code blocks; `\"x\"` escaping gone.
- `mix format --check-formatted lib/lua.ex` — clean.
- Website compiles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,10 +69,6 @@ rescue
     e.line    # => 2
     e.source  # => "<eval>" (chunk name)
 
-    # Render the message with Exception.message/1. The :message struct field
-    # is nil when the exception wraps a VM error — the frame is built lazily
-    # so ANSI color is gated on the terminal at output time. It's a formatted,
-    # colorized frame (ANSI codes elided here):
     #
     #   Lua runtime error: at <eval>:2:
     #
--- a/README.md
+++ b/README.md
@@ -69,14 +69,15 @@ rescue
     e.line    # => 2
     e.source  # => "<eval>" (chunk name)
 
-    # e.message is a formatted, colorized frame (ANSI codes elided here):
+    # Render the message with Exception.message/1. The :message struct field
+    # is nil when the exception wraps a VM error — the frame is built lazily
+    # so ANSI color is gated on the terminal at output time. It's a formatted,
+    # colorized frame (ANSI codes elided here):
     #
-    #   Lua runtime error: Runtime Error
-    #
-    #     at <eval>:2:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -69,14 +69,15 @@ rescue
     e.line    # => 2
     e.source  # => "<eval>" (chunk name)
 
-    # e.message is a formatted, colorized frame (ANSI codes elided here):
+    # Render the message with Exception.message/1. The :message struct field
+    # is nil when the exception wraps a VM error — the frame is built lazily
+    # so ANSI color is gated on the terminal at output time. It's a formatted,
+    # colorized frame (ANSI codes elided here):
```
davydog187:
```suggestion
```

PR #393, opened 2026-07-15
errors: make Exception.message/1 plain; add Lua.format_exception/1

davydog187:
## Context

A user reported that `Exception.message/1` on Lua's exceptions embeds ANSI color **and** a multi-line layout (location header, source context, stack trace, suggestion). When they start their app in dev from a terminal, `IO.ANSI.enabled?/0` is legitimately `true`, so every `Logger.error(Exception.message(e))` gets colored, multi-line log lines.

The root cause is a layering issue: `IO.ANSI.enabled?/0` answers *"is stdout a TTY?"*, a poor proxy for *"is this string about to be painted onto an interactive terminal?"*. The two come apart the moment the string goes to `Logger` instead of `IO.puts`. And ANSI is only the visible symptom — even with color off, the message was still multi-line.

Colored, rich rendering is only genuinely useful for a human reading an error interactively — i.e. `mix lua.eval` and uncaught crashes. So this PR moves it there and makes the library default plain.

## Changes

- **`Exception.message/1`** (on `Lua.RuntimeException` / `Lua.CompilerException`) now returns a **plain, single-line, ANSI-free** string — the error body plus a compact `(at source:line)` suffix — safe for `Logger` and error trackers.
- **`Lua.format_exception/1`** renders the rich, ANSI-gated report (location, stack trace, suggestions) and is what `mix lua.eval` now prints. Each internal VM error gained a `format/1` sibling to its `message/1` / `to_map/2`.
- **`Lua.RuntimeException.to_map/2`** and **`Lua.CompilerException.to_map/1`** expose the existing wire-safe structured shape (no ANSI) for JSON payloads, structured logs, and UI-facing reporting.

### Before / after (app started from a TTY, `ansi_enabled: true`)

```elixir
# before: colored + multi-line in the log
Logger.error(Exception.message(e))

# after: one clean line, no escapes
Exception.message(e)
#=> "Lua runtime error: attempt to perform arithmetic on a nil value (local 'x') (at demo.lua:2)"

Lua.format_exception(e)   # rich report, colored on a TTY — for mix lua.eval / interactive use
Lua.RuntimeException.to_map(e)   # %{message:, source:, line:, call_stack:, ...} — for UIs
```

## Safety: Lua §6.1 is untouched

The value `pcall`/`xpcall`/`Lua.call_function` hand back to Lua goes through `Lua.VM.ProtectedCall.error_value/1` (raw `:value` / `:lua_value` / `ArgumentError.raw_message`), **not** `Exception.message/1`. Verified: `pcall(error('boom'))` still returns `{false, "<eval>:1: boom"}`.

## Compatibility note

If you relied on `Exception.message/1` for the rich render, switch to `Lua.format_exception/1`. Message strings aren't a stable API, but this is called out in the CHANGELOG since the output format visibly changes.

## Verification

- `mix test` → **2611 passed**, 7 skipped; `mix format --check-formatted` clean; compiles with `--warnings-as-errors`.
- `mix lua.eval` on a failing script (piped/non-TTY): full rich report, **zero** ANSI escapes.
- Kevin's scenario (`ansi_enabled: true`): `Exception.message/1` has no `\e[` and no `\n`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,32 +47,21 @@ Everything else — the default sandbox, `_G`/`_ENV` semantics, metatables, and
 the standard-library surface — is compatible. The full breaking-change list
 is in the [`1.0.0-rc.0`](#100-rc0---2026-05-26) entry below.
 
-## [Unreleased]
-
-### Added
-- `Lua.format_exception/1` renders a `Lua.RuntimeException` or
-  `Lua.CompilerException` as the rich, human-readable report — location, source
-  context, stack trace, and suggestions — with ANSI color when
-  `IO.ANSI.enabled?/0` is true. This is the report `mix lua.eval` prints.
-- `Lua.RuntimeException.to_map/2` and `Lua.CompilerException.to_map/1` expose a
-  wire-safe structured representation (no ANSI) for JSON payloads, structured
-  logs, and UI-facing error reporting (#393).
-
-### Changed
-- `Exception.message/1` on `Lua.RuntimeException` and `Lua.CompilerException`
-  now returns a plain, single-line, ANSI-free string suitable for `Logger` and
-  error trackers, rather than the multi-line ANSI report. The rich report moved
-  to `Lua.format_exception/1`. This removes ANSI escapes from logs even when the
-  app is started from a TTY (where `IO.ANSI.enabled?/0` is true). If you relied
-  on `Exception.message/1` for the rich render, switch to `Lua.format_exception/1`
-  (#393).
-
 ## [1.0.0] - 2026-07-15
 
 The first stable release on the Elixir-native Lua 5.3 VM, culminating the
 `1.0.0-rc.0` through `rc.3` series. Upgrading from the last public release,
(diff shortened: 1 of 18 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,26 @@ Everything else — the default sandbox, `_G`/`_ENV` semantics, metatables, and
 the standard-library surface — is compatible. The full breaking-change list
 is in the [`1.0.0-rc.0`](#100-rc0---2026-05-26) entry below.
 
+## [Unreleased]
```
davydog187: this should go under the 1.0.0 section. we are releasing this as part of 1.0.0

davydog187: Done in caa4c50 — moved the `format_exception` / `to_map` and `Exception.message/1` entries out of `[Unreleased]` and into the `1.0.0` section. Also reconciled the two existing `#384` entries in that section, since `#393` supersedes them (message is now plain; the rich report moved to `Lua.format_exception/1`), so the release reads as one coherent story instead of contradicting itself.

PR #406, opened 2026-07-27
fix(pattern): honour ^ anchor in gsub and literal caret in gmatch

smn:
## What

Fix two related `string` pattern bugs where a leading `^` was mishandled:

- **`string.gsub` ignored the `^` anchor.** An anchored pattern only
  matches at the start of the subject (Lua 5.3 §6.4.1), so `gsub` must
  attempt at most one replacement at position 0 and leave the remainder
  untouched. We were discarding the `anchored` flag from `compile/1` and
  scanning every position, so `("hello"):gsub("^h", "H")`-style calls
  replaced interior matches too.

- **`string.gmatch` treated a leading `^` as an anchor.** In `gmatch` a
  leading `^` has no special meaning (Lua 5.3 §6.4.3) — PUC-Lua's
  `gmatch_aux` never strips it, so `match` sees it as an ordinary
  character. We were honouring it as an anchor, which prevented iteration.

## How

- `gsub`/`gsub_stateful` now route through a new `do_gsub/6` that
  branches on the `anchored` flag:
  - anchored → match once at position 0, splice the replacement onto the
    untouched remainder, return a count of 0 or 1. `max_n <= 0` short-
    circuits to zero replacements.
  - unanchored → existing `gsub_from/10` full scan.
- `gmatch` recompiles the pattern with the caret kept literal when
  `compile/1` reports it as anchored, mirroring PUC-Lua's behaviour.

## Tests

New `test/lua/vm/stdlib/pattern_anchor_test.exs` covering:

- **Anchored `string.gsub`** — leading-only replacement, no match when
  the subject doesn't start with the pattern, multi-char runs replaced at
  most once, empty anchored match, leading-whitespace trim preserving
  interior/trailing whitespace, `n = 0` suppression, captures reaching a
  function replacement exactly once, and whole-subject matches.
- **Leading caret in `string.gmatch`** — `^` does not anchor and matches
  only as a literal character.
- **Anchored `Pattern.gsub/4`** — at-most-once replacement and `max_n = 0`.

### DIFF
--- a/lib/lua/vm/stdlib/pattern.ex
+++ b/lib/lua/vm/stdlib/pattern.ex
@@ -74,7 +74,16 @@ defmodule Lua.VM.Stdlib.Pattern do
   Global match - returns list of all matches as {start, stop, captures}.
   """
   def gmatch(subject, pattern) do
-    {_anchored, pattern_elems} = compile(pattern)
+    # Lua 5.3 §6.4.3: in gmatch a leading `^` does not work as an anchor
+    # (it would prevent the iteration). PUC-Lua's gmatch_aux never strips
+    # it, so `match` sees it as an ordinary character — recompile the
+    # pattern with the caret kept as a literal.
+    pattern_elems =
+      case compile(pattern) do
+        {false, elems} -> elems
+        {true, _elems} -> compile_elements(pattern, [])
+      end
+
     gmatch_from(subject, 0, byte_size(subject), pattern_elems, [], -1)
   end
 
@@ -121,7 +130,7 @@ defmodule Lua.VM.Stdlib.Pattern do
   `gsub_stateful/5` instead.
   """
   def gsub(subject, pattern, repl, max_n \\ nil) do
-    {_anchored, pattern_elems} = compile(pattern)
+    {anchored, pattern_elems} = compile(pattern)
 
     stateful_repl =
       if is_function(repl, 1) do
@@ -130,8 +139,7 @@ defmodule Lua.VM.Stdlib.Pattern do
         repl
       end
 
-    {result, count, _state} =
-      gsub_from(subject, 0, byte_size(subject), pattern_elems, stateful_repl, max_n, 0, [], nil, false)
+    {result, count, _state} = do_gsub(subject, anchored, pattern_elems, stateful_repl, max_n, nil)
 
     {result, count}
   end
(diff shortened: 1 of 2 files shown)


davydog187:

Two independent AI reviewers (Opus, Sonnet) reviewed this PR, both by differential testing against a real Lua interpreter. **The fix is correct**: a 10,200-case differential fuzz (20 subjects × 34 `^`-leading patterns × string/function/table replacements × `n` ∈ {nil, 0, 1, 2, −1} × `gsub`/`gmatch`/`find`/`match`) produced **0 divergences from PUC-Lua** on this branch, vs 3,224 divergences on pre-fix `main` — so both bugs are real and the fix is complete over that space. Error/state-threading probes (callback errors, side-effect preservation, nested gsub in an anchored callback, `%2` invalid index) all match PUC too. Semantics were cross-checked against `lstrlib.c` (`str_gsub`'s `if (anchor) break;`, `gmatch_aux` never stripping `^`). Thanks @smn — nice fix.

Everything below is take-it-or-leave-it polish, no blockers:

### 🟡 LOW

- `lib/lua/vm/stdlib/pattern.ex:77` (and `pattern_anchor_test.exs:8`) — the comment cites "Lua 5.3 §6.4.3", which doesn't exist (5.3 has §6.4.1 Patterns and §6.4.2 Pack format); the rule lives in the `string.gmatch` entry of §6.4. Also, the "prevented iteration" framing is backwards: pre-fix `gmatch("aaa","^a")` silently dropped the anchor and *over-matched* (3 unanchored matches), rather than preventing iteration.
- `lib/lua/vm/stdlib/pattern.ex:81-85` — `gmatch` compiles a `^`-leading pattern twice (`compile/1`'s result is discarded, then `compile_elements(pattern, [])` re-parses). A `String.starts_with?(pattern, "^")` check would express the intent without the wasted parse. (Both reviewers flagged this independently.)
- `lib/lua/vm/stdlib/pattern.ex:169-180` — the anchored `do_gsub` clause re-implements the match → `apply_replacement` → splice sequence `gsub_from/10` already owns; verified equivalent today, but a future `gsub_from` fix (e.g. the `%f` frontier work) would silently miss the anchored path.
- `lib/lua/vm/stdlib/pattern.ex:165` — the `max_n != nil and max_n <= 0` guard passes non-numeric `max_n` via term ordering (reachable only through the Elixir-level `Pattern.gsub/4`; `string.gsub` coerces first). `is_integer(max_n) and max_n <= 0`, applied once before the anchored/unanchored split, would be tighter.
- Style: the positional-boolean dispatch `do_gsub(subject, true/false, …)` would read better as `gsub_anchored/gsub_unanchored` (or `:anchored`/`:unanchored` atoms).

### Suggested test additions (all verified to behave correctly already — just unpinned)

- Anchored `gsub` with `n > 1`: `("aaa"):gsub("^a", "X", 3) == "Xaa", 1` — this is the test that pins "a positive limit still allows at most one replacement", the core claim of the fix.
- Anchored `gsub` with a **table** replacement (that branch never sees the anchored path in any test) and with a failing `$` (`("abc\n"):gsub("^%a*$", "X")` → unchanged, 0).
- `gmatch` literal-caret beyond the plain case: captures (`("^a^b"):gmatch("^(%a)")` → `a`, `b`), a quantifier bound to the caret, caret-plus-`$`.
- Worth noting: the vendored suite's anchored cases (`pm.lua:95/:148/:161`) passed *before* this fix by accident (unanchored scan happens to match at position 0 for the trim idiom), so this PR's unit file is the real guard.

### Out of scope (observed while testing; will be filed as issues, not this PR's problem)

- `print()` of a non-UTF-8 binary raises (`:io.put_chars` argument error) where PUC writes raw bytes — e.g. `print(("日本"):gsub("^.", "X"))` crashes even though the gsub result is byte-identical to PUC.
- `"invalid capture index %2 in replacement string"` is Lua 5.4 wording; PUC 5.3 says `"invalid capture index %2"`.

davydog187:
Thanks for this — nice catch, and the diagnosis is exactly right. I ran a differential fuzz of the patched `gsub`/`gmatch` paths against PUC-Lua (10,200 cases) and got 0 divergences, so the fix is good as-is.

I pushed one test-only commit (`e2ddae1`) to the branch, filling in a few anchor cases that weren't pinned yet:

- anchored `gsub` with **`n > 1`** — the most important one, since it's the core claim of the fix and the file only covered `n = 0`. `("aaa"):gsub("^a", "X", 3)` -> `"Xaa", 1`.
- anchored `gsub` with a negative `n`, and with a table replacement (the table branch never reached the anchored path in any test)
- an anchored pattern whose trailing `$` fails
- `gmatch` literal caret with a capture, with a quantifier bound to the caret, and followed by an end anchor

All expected values were taken from the reference interpreter rather than written by hand. I also corrected the `§6.4.3` citation in the comments — 5.3 has no such section; the `gmatch` caret rule lives in the `string.gmatch` entry of §6.4.

No `lib/` behaviour changed. Feel free to squash or reword that commit however you like.

smn:
awesome, thank you!

PR #407, opened 2026-07-27
Fix unbounded recursion on cyclic tables at the eval boundary

fedme:
## The problem

The most common Lua OOP idiom creates a table that contains itself:

```lua
local Client = {}
Client.__index = Client   -- Client's own "__index" entry points back at Client

function Client.new(opts)
  return setmetatable({}, Client)
end

return Client
```

`Client.__index = Client` is how instances delegate method lookup to the module table, so virtually every hand-rolled Lua "class" has this shape. Returning such a table from `Lua.eval!/3` never terminates on 1.0.1. Both eval-boundary walks recurse into the cycle and allocate a fresh snapshot of the whole table at every level:

* `decode: true` (the default): `Lua.VM.Value.decode/2` recursing through the `{:tref, id}`
* `decode: false`: `Lua.VM.Display.peek_table` building the peek for `Display.Table`

Memory grows until something dies. We hit this in production hosting sandboxed Lua apps: worker processes run with `max_heap_size` set, so the BEAM killed them mid-walk with an untrappable `:killed` while loading any module using this idiom (observed climbing past 2.9GB uncapped). The minimal reproduction is three lines:

```lua
local T = {}
T.__index = T
return T
```

## The fix

Thread a set of ancestor table ids through both walks and stop at the point of recurrence:

* **`decode: true`**: the walk terminates by leaving the table's `{:tref, id}` reference at the recurrence. This mirrors the existing contract, where functions already pass through decoded tables as opaque references; a cyclic table is just another value with no acyclic plain-data representation. No new option and no behavior change for acyclic data.
* **`decode: false`**: the recurring occurrence carries `peek: :circular` and renders as `#Lua.Table<id: N, circular>`.

The ancestor set is path-based, so shared references that do not form a cycle (the same table under sibling keys) still decode and peek in full. Tests cover self-cycles, mutual cycles, shared non-cyclic references, and the sibling-key false-positive case.

One design note worth flagging: with reference substitution, someone serializing decoded data discovers a cycle at encode time (`cannot encode {:tref, 12}`) rather than at the eval boundary. The alternative is raising a clear error when decode meets a cycle. We went with substitution for consistency with function passthrough, but happy to switch if you prefer the strict version.

### DIFF
--- a/lib/lua/vm/display.ex
+++ b/lib/lua/vm/display.ex
@@ -93,7 +93,7 @@ defmodule Lua.VM.Display do
   see `Lua.VM.Display.Table`.
   """
   @spec wrap_value(term(), State.t(), boolean()) :: term()
-  def wrap_value(value, state, decode?), do: wrap_value(value, state, decode?, MapSet.new())
+  def wrap_value(value, state, decode?), do: wrap_value(value, state, decode?, %{})
 
   # decode: true — only wrap closures/native; tables and userdata
   # have already been decoded and are passed through unchanged.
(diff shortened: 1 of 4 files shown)


fedme:
@smn this is the issue I mentioned

davydog187:
Thanks! 

🍍 🍍 🍍 🍍 🍍

fedme:
> Thanks!
> 
> 🍍 🍍 🍍 🍍 🍍

Thank you for reviewing it! I've switched `MapSet` to a plain `Map` because Dialyzer was complaining about it.

davydog187 (approved):

PR #418, opened 2026-07-28
fix(parser): allow comments between a bare return and its terminator

smn:
## Problem

A valueless `return` followed by a **comment** before the block terminator (`elseif` / `else` / `end` / `until`) fails to parse:

```lua
local function f(method)
  if method == "GET" then
    return
  -- handle other methods
  elseif method == "POST" then
    return
  end
end
```

```
Parse Error: Expected expression
```

Comments are whitespace in Lua, so the `return` here is simply a bare (valueless) return and the branch continues normally. Reference Lua 5.3 accepts this, and it's a common idiom (`then return -- note \n elseif ...`).

## Root cause

`Lua.Parser.parse_return/1` peeked at the **raw** next token to decide whether the `return` was valueless. A comment token is not one of the terminators (`end`/`else`/`elseif`/`until`), nor EOF/`;`, so it fell through to `parse_expr_list`, which then choked on the comment → "Expected expression".

## Fix

Peek **past** comment tokens (via the existing `skip_comments/1`) when detecting the terminator. The comment tokens are left in the stream so the block parser still collects them as orphaned/trailing comments — no comment metadata is lost.

```elixir
case peek(skip_comments(rest)) do
```

The semicolon branch consumes past comments as well, so `return -- note` followed by `;` is handled too.

## Tests

Added a regression test in `test/lua/parser/statement_test.exs` covering a bare `return` with a comment before `elseif`/`else`/`end` — fails before, passes after.

- `mix test test/lua/parser/` — all green
- `mix test` (full suite) — all green

### DIFF
--- a/lib/lua/parser.ex
+++ b/lib/lua/parser.ex
@@ -252,7 +252,11 @@ defmodule Lua.Parser do
 
   # Placeholder implementations for statements (Phase 3)
   defp parse_return([{:keyword, :return, pos} | rest]) do
-    case peek(rest) do
+    # Comments are whitespace between the `return` keyword and the block
+    # terminator, so look past them when deciding whether this is a bare
+    # (valueless) return. For terminator/EOF the comment tokens stay in `rest`
+    # for the block parser to collect as orphaned/trailing comments.
+    case peek(skip_comments(rest)) do
       # End of block or statement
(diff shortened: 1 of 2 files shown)


davydog187 (approved):

davydog187:
Thanks!

🍍 🍍 🍍 🍍 🍍

PR #419, opened 2026-07-28
release: 1.0.2

davydog187:
Prepares the `1.0.2` release: bumps `@version` in `mix.exs` and adds the CHANGELOG entry for everything merged since `v1.0.1`.

Patch bump — all fixes and internal performance work, no public API change.

### Changed
- `Lua.new/1` ~100x faster via a memoized boot-time VM template (#398)
- VM dispatcher / call-convention overhead cuts (#405, #401)
- Compiler peephole pass + more work done once at compile time (#403, #416, #400, #399)

### Fixed
- Comments between a bare `return` and its terminator now parse (#418)
- `^` anchor honoured in `string.gsub`; literal caret in `string.gmatch` (#406)
- No more unbounded recursion on cyclic tables at the eval boundary (#407)

The website-only playground change (#402) is intentionally omitted — it doesn't affect the published package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,8 @@ is in the [`1.0.0-rc.0`](#100-rc0---2026-05-26) entry below.
 
 ## [Unreleased]
 
+## [1.0.2] - 2026-07-28
+
 ### Changed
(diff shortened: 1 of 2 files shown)


davydog187:
Separately i'm running benchmarks to verify the 100x claim

PR #422, opened 2026-09-01
fix(vm): boolean table keys, nan sentinel, set! encode errors, # on non-tables

### LINKED ISSUE #404: length operator on nil/boolean returns 0 instead of raising
Found while building an error-fidelity test battery for the perf work (#403).

`return #nil` and `return #true` both evaluate to `0`. PUC-Lua 5.3 raises:

```
lua: attempt to get length of a nil value
lua: attempt to get length of a boolean value
```

Present on main independent of the peephole pass (reproduces with `peephole: false` and on commits before #403). Likely a missing type check in the length-operator implementation shared by both engines.

dimamik:
Thanks a lot for building this! We're about to release [Legion](https://github.com/software-mansion-labs/legion), one of the sandboxes of which will be Lua.ex, and we stumbled upon a few non-urgent thingies around data transformation that we think are worth looking into:

- `Value.encode/3` turned `true`/`false` map keys into strings, so `%{true => 1}` came back as `{["true"] = 1}`.
- The `:nan` sentinel was not a number at the boundary: `type(0/0)` was `"userdata"`, `math.type(0/0)` was `nil`, `tostring(0/0)` was `":nan"`, and `Lua.decode!/2` raised on it.
- `Lua.set!/3` crashed with a bare `FunctionClauseError` on values `Lua.encode!/2` refuses with `Failed to encode`.
- `#` returned 0 for nil, booleans, numbers and functions instead of raising. Fixes #404.

Thanks in advance, and let me know what you think!


dimamik:
@davydog187 - Could I ask you to take a look? Thanks!!!

